PR #2, opened 2015-02-04
Convert to Markdown

TheLastProject:
README converted to Markdown to make it show up more prettily on Github.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,4 +1,4 @@
-== Red Eclipse ==
+# Red Eclipse
 * A free, casual arena shooter; released as Open Source
 * Available for Windows, GNU/Linux, BSD, and OS X
 * Features parkour, impulse boosts, dashing, and other tricks
@@ -6,7 +6,7 @@
 * Builtin editor lets you create your own maps cooperatively online
 * Free download from http://redeclipse.net/
 
-== About ==
+## About
 Red Eclipse is a fun-filled new take on the casual first person arena shooter, built as a total conversion of Cube Engine 2, which lends itself toward a balanced gameplay, with a general theme of agility in a variety of environments. For more information, please see our Wiki or Forum.
 
 Wiki: http://redeclipse.net/wiki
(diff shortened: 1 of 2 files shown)


qreeves:
I'm not sure it is a great idea to rename the file from .txt to .md. It has been this way for a long time so Windows users can double click the file to open it in their text editor. I know that sound silly, but I think it is important.

readme.md
```diff
@@ -77,12 +77,12 @@ The project is Free and Open Source meaning that you can both use it for free, a
 * Rob "Lloir" Shannon
 * Robert "Homicidal" Crane
 * Robert Winkler
-* Ruben "TheLastProject" van Os - Design, Testing and Support
+* "TheLastProject" - Design, Testing and Support
 * Ryan "icculus" Gordon / icculus.org - Master server and website hosting
-* R�mi "LuckystrikeRx" Clouet d'Orval - Maps, Textures, Models, Weapons, Design/Testing, Website
+* Rémi "LuckystrikeRx" Clouet d'Orval - Maps, Textures, Models, Weapons, Design/Testing, Website
```
qreeves: Also, why did this change? :stuck_out_tongue:

TheLastProject: Github's editor changed, as it is displayed in vim for me, <E9> to é, so probably a different character encoding or so. There's no other change there.

TheLastProject:
Valid point. I have restored readme.txt. This does mean both readme.md and readme.txt will have to be updated in case anything is to be changed, but seeing how the README is unlikely to be updated often anyway this doesn't seem like a serious problem to me.

ienorand:
It is apparently possible to use a symlink (README.md -> readme.txt) in order to force parsing of any file on GitHub (symlinks in Git ends up as dummy files on windows, so won't cause issues).

qreeves:
OK, yeah. Do what @ienorand said. Doesn't matter if the txt is in md format.

PR #27, opened 2015-02-05
C++11 HURRAY!

FaTony:
It been good 4 years for C++ and compiler vendors to produce production quality c++11 compilers. G++ 4.8.1, Clang 3.5 and other offset full or almost full support for c++11 and some even c++14. Two years ago, when I first proposed moving to c++11, it could've been to hasty but now is the right time.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -3,7 +3,7 @@ APPCLIENT=$(APPNAME)
 APPSERVER=$(APPNAME)_server
 
 #CXXFLAGS= -ggdb3
-CXXFLAGS= -O3 -fomit-frame-pointer -ffast-math
+CXXFLAGS= -std=c++11 -O3 -fomit-frame-pointer -ffast-math
 override CXXFLAGS+= -Wall -fsigned-char -fno-exceptions -fno-rtti
 
 PLATFORM= $(shell uname -s)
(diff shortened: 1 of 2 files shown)


graphitemaster:
Having started a project with the primary objective to use C++11. My experience in the domain has taught me that support for C++11 is not well supported. While it may be true that clang and g++ both support C++11 almost entirely, their regression count for C++11 is huge (g++ has over 6 thousand C++11 bugs not yet fixed.). Other compilers like MSVC only support a scarce subset of it. Similarly speaking, most distributions currently (Ubuntu for instance) don't ship with the recent compilers like 4.8.1. I vote against this.

qreeves:
Yeah, I'd rather not risk the regressions. I know you're passionate about this subject, but it is us who has to maintain and deal with whatever issues it causes, so I have to err on the side of "If it isn't broken, don't fix it.". Sorry.

FaTony:
FYI RE builds and runs flawlessly under c++11.

qreeves:
I know, but I'm not versed enough in the topic to make this decision, and therefore have to defer to those on the project that are. Maybe you could do it as a separate build pipeline? Eg: "make c11" and leave the default the way it is, then it could be optional.

JeffBobbo:
There's no point in doing that, because the C++11 features won't be usable to anyone whose not compiling a C++11 version. The only advantage would be any back end changes that don't change interfaces we use, and given how little of the STL is actually used the change is pretty pointless.

PR #29, opened 2015-02-06
scoring help suggestion for shirepirate to review

bonifarz:
Should be equivalent to http://redeclipse.net/forum/viewtopic.php?f=4&t=231&start=50#p5126
Not exactly "finalized" work, but hopefully an improvement to the current revision.


shirepirate:
and the main difference here is?

bonifarz:
Difference to forums: Colored var lookups (got an older one here by mistake, fixed that).
Differences we discussed earlier: Team kill double penalty in ctf/dnc/bb, some menus rearranged/renamed. As usual: Just a suggestion. I've put it here in case that is more convenient for you to work with.

bonifarz:
It does not really matter that I deleted the branch unintentionally - I think the pull request here was not really useful for shirepirate anyway. I'm convinced that his final polish of the help menus will be better than any snippet I can provide. And thanks again for the good work.

shirepirate:
sorry, just hadn't gotten to looking at it again. it seems OK, no reason not to add it. the more accurate information the better

PR #30, opened 2015-02-06
usage descriptions merged with suggestions from the forums and wiki

bonifarz:
from topic http://redeclipse.net/forum/viewtopic.php?f=4&t=368
gui descriptions in accordance with http://redeclipse.net/wiki/CubeScript#Creating_menus


shirepirate:
what about mentioning the gui wrap for text as well? should be the last argument

bonifarz:
Ow, missed that guitext wrap arg. Will add it and a few more gui commands (blend, bg, slider etc). Just trying to figure out how guieditor works. In https://github.com/red-eclipse/base/blob/master/src/engine/menus.cpp we can see the argument names and their defaults, but unserstanding them all needs some more trial and error :)

qreeves:
I'm monitoring, so if in doubt, ask.

PR #31, opened 2015-02-07
a search filter for serverdesc and players

bonifarz:
A search field and checkbox to only show servers with matching player names, player handles or server descriptions. A proper implementation is certainly preferable, maybe via the sort functionality, but this hack seems to work well enough to be useful. So, I'm not sure if it is worth pushing.
Related topic: http://redeclipse.net/forum/viewtopic.php?f=5&t=490&p=4216

### DIFF
--- a/config/menus/servers.cfg
+++ b/config/menus/servers.cfg
@@ -137,10 +137,6 @@ servermenu = [
                 ]
             ]
         ]
-        guilist [
-            guifield searchtagval 20 [searchtag = $searchtagval; search = 1] -1 0 "" 0 "^fd <enter search terms>"
-            guicheckbox "only show servers with matching players or descriptions" search 
-        ]
         guistrut 0.25
         if (> (getversion 3) (getversion 1)) [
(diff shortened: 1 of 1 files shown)


qreeves:
I just checked this out, the presentation leaves a lot to be desired, so I suggest you clean up how it looks if you want this merged. We're still aiming to get 1.5 out soon, so I don't want a bunch of half finished ideas going into master.

Also, if you take a look at my [vars search](https://github.com/red-eclipse/base/blob/master/config/menus/vars.cfg#L14), I use a newly added feature to update the variable as it changes in the search field, which would also be useful here.

bonifarz:
Thanks for the feedback. 
I really did not mean to bother you, but simply could not think of a better way to present this: Keeping it compact vs making clear what it does. I just tried my best to get a better compromise for this. Or is the main issue that an (almost) empty server list looks ugly?

Either way, if the hack helps, I'm happy, if it is better to leave it as an unofficial modification, that is just fine for me. But again, I don't want to cause trouble with more work for you :)

Hey, I did not notice the new argument of guifield. That's a very cool feature!

PR #39, opened 2015-02-11
Windows devel caching

freezurbern:
I changed the scripts to use My Documents or Documents or finally the Red Eclipse directory because in my opinion, Red Eclipse shouldn't put things in AppData. AppData is just as messy as the Windows registry, and is yet another place to keep track of. I also included elevate.exe from the author's website, as it seemed to be missing in the repo (perhaps because of the gitignore settings)

With the cache folder in My Games instead of AppData, it is easier to develop as well as troubleshoot others' problems as AppData is by default hidden from the average user. Also, My Games should have the same permissions as AppData for the average user.


Evropi:
I agree with it, it increases RE's usability greatly. Many other games put themselves in the My Documents subfolder _My Games_ with good reason. Firstly, because multiple users on the same machine may play the game and don't want their profiles to overlap. Secondly, My Games makes it much easier to find and store downloaded maps or maps they have created themselves.

shirepirate:
well, we're not talking about profiles or game saves or even map files so much as an archive backup of the binaries, as i see it. i doubt it really matters, though i found if the binaries are deleted from /bin/ thereafter neither .bat file seems to be able to handle that properly, it bugs out.

qreeves:
Yeah, I'm working on getting the updater working before I make it "recursive" and retry if binary execurion fails.

PR #40, opened 2015-02-11
enhance search filter for server menu

bonifarz:
Continuation of https://github.com/red-eclipse/base/pull/31
Allows to search also for maps, modes, mutators (the full names) and ip (the latter seems a bit strange, but was requested here http://redeclipse.net/forum/viewtopic.php?f=5&t=490&p=4216#p4212)
A bit of cleanup, too.

### DIFF
--- a/config/menus/servers.cfg
+++ b/config/menus/servers.cfg
@@ -23,7 +23,6 @@ sinfomodify = [
     ]
     serversort $sinfotlist
 ]
-
 servermenuinit = [
     sinfoindex = 0
     sinfowait = 0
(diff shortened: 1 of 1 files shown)


shirepirate:
considering IPs are publicly displayed anyways i doubt it will be a problem.
the tooltip is a little wordy though, can you summarize it so it's easier to read and make sense of it?
something like "filter by server name or description, player name or auth handle, IP address, map name or current mode/mutator"

qreeves:
I agree with @shirepirate - once that's taken care of it can be merged.

bonifarz:
Thanks for the suggestion, @shirepirate
So, what about this one:
"Filter by player names or handles, current map, mode or mutators, server description, IP address or port."
I'm fine with any other tooltip.

PR #41, opened 2015-02-12
Show the previous value of a variable when it is changed.

### LINKED ISSUE #38: [enhancement] Vars changing display
(http://redeclipse.net/forum/viewtopic.php?f=4&t=520&p=4607&hilit=big+list#p4607)

If u change a variabel in the game it shows only to what it was changed, i think (mainly for coop vars editing) its usefull and a good option to show from what it was changed, so not only changed to but also the value which was changed. Still an information to the others.
example:

    Iceflower set timelimit to 0

i would like to see :D

    Iceflower set timelimit from 10 to 0

shacknetisp:
closes #38

### DIFF
--- a/src/game/client.cpp
+++ b/src/game/client.cpp
@@ -1119,7 +1119,7 @@ namespace client
             {
               if(oldval)
               {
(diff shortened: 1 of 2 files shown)


qreeves:
The code in this commit does not conform to the Cube source guidelines. Please use four (4) spaces for indentation.

PR #43, opened 2015-02-12
add experimental search for maps menu

bonifarz:
The custom map field can be used for searches (with a checkbox).
Matching maps are shown at the top (sort rather than filter).
Experimental, so maybe the implementation is rather inefficient.
I'll just put this here for review, in case it is useful.


bonifarz:
Thanks for your time and for the useful feedback. I'll take this into account when I will fix a silly mistake that results in duplicate matches *facepalm*

bonifarz:
I realize it is a bit more complicated to do this right, so I better close this pull request as postponed. Once I have something that is finalized, we can still open a new request.

Issue #13, opened 2015-02-05
Translate IRC colours to in-game


qreeves:
From: http://redeclipse.net/forum/viewtopic.php?f=9&t=460

Since Red Eclipse already translates in-game colors to IRC colors, as per a var in the servinit.cfg, I would very much like to see the reverse as well. That way, color codes are not shown in irc-to-game output.

ZeroKnight:
I could be unclear on exactly what the request is, but IRC->Cube color codes already work, albeit not entirely as they should. [This line](https://github.com/red-eclipse/base/blob/master/src/engine/irc.cpp#L209) shows that all it does is find any EOT escapes (`\x03` / `^C`) that IRC clients use as the start of a "color code" and converts it to a FF (`\f`), which is the Cube Engine equivalent.

However, the colors that each number in the color escape sequence represents may not match the ones that Cube uses depending on the IRC client. This is because the IRC protocol does not in itself support color codes and is a feature implemented by the IRC client, thus the color mappings may vary, or not exist at all. Most clients that I know of conform to [mIRC color codes](http://www.mirc.com/colors.html).

There are two problems with the current implementation:

* Cube color codes will not match up with most (if not all) IRC clients
* 2-digit IRC color codes will not work

Regarding the first point, examine the table below. What appears as say, *Green* in a message sent in IRC would appear as *Red* in-game, because the color code **3** is *Green* in IRC, but *Red* in Cube.

Code | mIRC | Cube
--------|---------|--------
0 | White | Green
1 | Black | Blue
2 | Blue | Yellow
3 | Green | Red
4 | Red | Grey
5 | Brown | Magenta
6 | Purple | Orange
7 | Orange | White
8 | Yellow  | Black
9 | Light Green | Cyan

Regarding the second problem, Cube only uses a single character to determine the color in the color escape sequence (not counting the use of brackets), where IRC color codes can use 2 digits. Consider the following message, sent by a user in IRC:

    ^C13this message is in pink!

In IRC, it would display: `this message is in pink!`, in pink/magenta as intended. However, in game chat you would see: `3this message is in pink!` in *Blue*. Notice the 3 at the beginning?

![irc-color-demo1](https://cloud.githubusercontent.com/assets/719733/6181823/1bcf883c-b307-11e4-85c5-950a1fda7aaa.png)

This is because only the `1` in `13` is a part of the Cube color escape sequence; the 3 is left as a part of the regular text and it is rendered Blue.

[...]

PR #47, opened 2015-02-13
Fix/Enhance IRC Relay colors. Fixes issue #13

ZeroKnight:
- Conforms to mIRC color codes, so most IRC clients should support them
  correctly.
- Normal, white colored Game->IRC text is no longer sent as pure black, but
  simply with no color code so IRC clients will use their default color. This
  prevents relayed game text being unreadable in IRC clients with dark
  backgrounds.

### DIFF
--- a/src/engine/irc.cpp
+++ b/src/engine/irc.cpp
@@ -116,7 +116,7 @@ void ircsend(ircnet *n, const char *msg, ...)
     }
 }
 
-void converttext(char *dst, const char *src)
+void cube2irc(char *dst, const char *src)
 {
     int colorpos = 0; char colorstack[10];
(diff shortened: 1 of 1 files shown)


src/engine/irc.cpp
```diff
@@ -116,7 +116,7 @@ void ircsend(ircnet *n, const char *msg, ...)
     }
 }
 
-void converttext(char *dst, const char *src)
+void converttext_cube2irc(char *dst, const char *src)
```
qreeves: How about naming these just `cube2irc` and `irc2cube`.

ZeroKnight:
>How about naming these just cube2irc and irc2cube?
@qreeves Done

PR #54, opened 2015-02-16
Readme updates

freezurbern:
updated the readme with some reorganization of credits, and gitattributes for windows friendlier things.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,8 +2,8 @@
 *.c     text
 *.h     text
 *.cpp   text
-*.txt   text eol=crlf
-*.cfg   text eol=crlf
+*.txt   text
+*.cfg   text
 *.bat   text eol=crlf
 *.sh    text eol=lf
 *.png   binary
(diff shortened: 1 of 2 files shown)


.gitattributes
```diff
@@ -2,8 +2,8 @@
 *.c     text
 *.h     text
 *.cpp   text
-*.txt   text
-*.cfg   text
+*.txt   text eol=crlf
+*.cfg   text eol=crlf
```
qreeves: These files shouldn't be in crlf format on non-Windows machines, this would do just that. They are correct in using the `text=auto` default as far as I know.

freezurbern:
@qreeves How should the files be handled then, if users can download
https://github.com/red-eclipse/base/archive/master.zip to get the game?
> (08:59:48)<angeles>anyway, as of now, windows users should be able to download the zip from github from the base repository
> (09:00:09)<angeles>it will consider it a stable version and keep it up to date

qreeves:
I can't seem to get it to fix them automatically, and both options are equally annoying for users of the opposing side of each. Perhaps all text files should be converted into markdown files, with some sort of auto-generation of HTML from that. I suggest just removing the git attribute changes for now so we can close this PR with the updates you provided, we'll figure it out later.

Issue #20, opened 2015-02-05
Targeted messaging


qreeves:
From: http://redeclipse.net/forum/viewtopic.php?f=9&t=8

It would be nice to be able to send messages to players. For example `/msg PLAYERNAME message` or you could add ID's which starts from 0. So it will be easier to send message to someone (`/msg ID text`).

ZeroKnight:
Don't forget to close this :)

PR #55, opened 2015-02-16
Targeted messaging

ZeroKnight:
Alright, here we are: whispers (aka: targeted/direct messaging) in Red Eclipse!

re-privmsg-ex

Adds 2 new commands: (closes issue #20)

    whisper <target> <text>     // send a message directly to another player
    mewhisper <target> <text>   // same as above, but in the form of an ACTION

As requested, the `N_TEXT` packet now takes a *target* argument as well, which as of right now is only used for whispers, -1 otherwise. This had the curiously helpful side effect of allowing whispers to show up in demos as also requested with no extra changes.

The exact formatting of the whisper in chat is not final (see later messages in discussion).

Also, I'd like to add a default alias for whisper to be `msg`, but my CubeScript is so rusty I can't remember how to go about doing an "all arguments" expansion. :sweat_smile: eg:

    alias msg [ whisper <all arguments> ]

Either way, I'm also not sure where it would go. `config/setup.cfg`, right with the saytextcommand stuff?


src/game/client.cpp
```diff
@@ -1016,20 +1016,28 @@ namespace client
...
-        defformatstring(m)("%s", game::colourname(d));
-        if(flags&SAY_TEAM)
+        defformatstring(n)("%s", game::colourname(d));
+        string s;
+        if(flags&SAY_PM)
         {
-            defformatstring(t)(" (to team %s)", game::colourteam(d->team));
-            concatstring(m, t);
+            defformatstring(pn)("%s", game::colourname(p));
+            formatstring(s)("\fc(MSG) %s -> %s\fw: %s", n, pn, msg);
```
qreeves: I don't know if I like the format of this, perhaps something similar without the need to prepend `(MSG)` on it?

ZeroKnight:
Oh, almost forgot. If you don't like how I formatted the private message, by all means throw down some suggestions, that's the best I could think up.

Lastly, does anyone have an opinion on a separate chat sound for private messages? The idea being you have an audible cue on top of a visual one... could make private messages *that* much more noticeable :)

src/game/server.cpp
```diff
@@ -5574,7 +5574,8 @@ namespace server
 
                 case N_TEXT:
                 {
-                    int lcn = getint(p), flags = getint(p);
+                    int lcn = getint(p), flags = getint(p), tcn = -1;
+                    if(flags&SAY_PM) tcn = getint(p);
```
qreeves: You should just make this part of the core N_TEXT message (eg. remove the conditional). Make it: `<from cn> <to cn> <flags> <text>` (if 'to' isn't set, use "-1"). Be sure to make every N_TEXT reference follow this procedure.

freezurbern:
@ZeroKnight I think a there should be a separate sound, because some people already don't notice team messages. Because private messages are (atleast by me,) deemed more important than regular chat or team messages, I think the sound shouldn't be subtle.

TheLastProject:
What about a simple coloured "From" and "To" followed by [icon] <username>: <message> for the private message look? Displaying your own nick seems quite redundant.

qreeves:
Yes, a new "ping" sound for PM's would be good. I can do that part if necessary.

Evropi:
Will this work over the IRC link as well?

ZeroKnight:
@Evropi As it stands, no it would not. Currently the only changes are to the game's network packet for text messages (`N_TEXT`) and is thus only inter-player on the same server.

qreeves:
While the messages aren't relayed to IRC, they should be stored in the demo, so if abuse occurs it can be seen when the demo is played back. Of course, this would require @ZeroKnight adjust the display function to show PM's not aimed at the player specifically during demo playback.

shirepirate:
in my opinion if anyone can view the messages through demo playback that provides a false sense of security which ought to inhibit the use of the phrase "private messaging." it's really very little better than some existing scripts for PMing

freezurbern:
@shirepirate perhaps it should be called direct messaging, instead then?

ZeroKnight:
@freezurbern Direct messaging seems a better fit IMO. :+1:

TheLastProject:
I don't quite agree that private/direct messages should be visible to others in demos.

ZeroKnight:
>@qreeves
>Did you consider supporting SAY_ACTION too?
Slipped my mind. Should be pretty simple to work in.

>@qreeves
>You should just make this part of the core N_TEXT message (eg. remove the conditional). Make it: `<from cn> <to cn> <flags> <text>` (if 'to' isn't set, use "-1"). Be sure to make every N_TEXT reference follow this procedure.
I did consider doing that, but now that I know you'd prefer it I can certainly change it up. It would probably look nicer than what I have now anyway.

qreeves:
@TheLastProject Quite often we are forced to choose between privacy and protection. It has always been the stance of this project to favour the latter (this same discussion happened with both IP addresses and team chat). You are using a public service, provided by an anonymous third party in most cases; you should not be using it for any personal/secure communications. By having them in the demo, they aren't available immediately, and provide accountability for moderators needing to review player behaviour (someone could abuse PM's by sending stuff that violates the guidelines and need dealing with)

TheLastProject:
@qreeves I understand, but I still don't feel it's a sane solution. It is misleading for a private message to be literally readable by any player. If we have to do something like this, which I still don't think is a good idea at all, I think the ability to view the private messages should at least be limited to moderators, not every random player on the server.

TheLastProject:
Seeing how @qreeves seems to be afraid that people will abuse private messaging to spam or otherwise break the guidelines, and wants to make them visible to everyone in a demo, I think we should just drop this feature.

Making the messages visible to everyone breaks the point of private messaging, and still doesn't help moderation because we'd need to get out of the game and watch a demo to figure out if someone was actually misbehaving, and have no way of dealing with them during the game.

This leads me to the conclusion that PMs, in this design, add no useful functionality, yet harm moderation, and we are better off telling people to just type the name of the player whose attention they want in the public chat.

src/game/client.cpp
Jigoku: As per formatting, why not use an icon:
```formatstring(s)("\fc\f(textures/chat) %s -> %s\fw: %s", n, pn, msg);```
Along with a colour that highlights the talk icon/emblem to (maybe have a user setting to change the colour of the icon for pm's)

bonifarz:
First: This looks awesome so far, @ZeroKnight 

Privacy: I guess direct chat does not have to be *pivate* in a strict sense to be useful, e.g. for a line of foreign language chat, or maybe for a tactical remark?

Abuse: We already have an /ignore command, right? I never used it, though. Maybe it will be useful to also have a variable for clients to enable or disable the proposed /msg feature. So, if you send a /msg to a client that has this disabled, you get a notification that this player does not want to receive any direct messages.

Evropi:
I agree on privacy that it should not be stored in demos. Demos are shared all the time. At least we should be able to strip private messages somehow.

Honno:
I'm with @TheLastProject about how to take advantage of this excellent opportunity @ZeroKnight has given to the game. If there is abuse of the system by one user targeting another, that user can simply chose to ignore the abuser via the /ignore command.

I understand that the integrity of RE is important, so "targeted messaging" would be OK as long as this isn't described as private messaging but indeed targeted messaging or something to that name. One suggestion is that when you type "/msg" into the games chatbox console thingy, the text that shows above (yaknow, where the syntax, default, max and mins are shown) could have a warning like "NOTE: This is not private and is potentially accessible by anyone".

Duckdef:
This is a very nice idea but I too am with @TheLastProject 
As mentioned, messages won't really be private - it just takes longer to access. 

I think it can be fixed by having another style of demo. The server owner should be able to determine which level of authority (mod, op, admin) can access this master demo. A simple command like /getmasterdemo can serve.

I agree with @qreeves on how its either privacy or protection, I believe a balance of it can work.

~Duck

qreeves:
And if there isn't a moderator around to grab the demo for the abused person? For it to be any use it is either/or, the ideas are mutually exclusive sorry.

The problem with `/ignore` is that it can be evaded quite easily by reconnecting to the server (it is client number based, so that'd need fixing too), and you are shifting the policing from the moderators to the user. If someone is being abusive enough to trigger this scenario, it is likely they should be banned, not ignored, lest they believe they can continue such behaviour (and teach others they can behave the same way). Even if ignore were fixed, you're expecting the user to know how to deal with the issue, which is not true in most cases (forget the whole language gap).

I've tried to explain to those here who are demanding the whole idea gets dropped if I insist on these safeguards (which I think that is a little selfish of them to be honest, they aren't in the position of having to deal with what I deal with); it has always been the policy of this project to favour protection and enforcement. Nothing you transmit to a third party server should be considered private or secure, in open source especially I have no way of ensuring it.

This is just an extension of the existing guidelines, you are in a public venue which is being recorded, there are plenty of other options outside of RE if you want a private conversation with someone. If you don't like that, don't use it. I'm not soliciting feedback, this is the way it is, end of story. I'm the one who ultimately ends up having to deal with the consequences of abuse, so it is my decision to make being abusive as difficult to hide as possible, and I don't make that decision lightly.

I know the pros and cons already, but my job is to make a game, not a private communication service. Please remember that.

bonifarz:
I fully agree with @qreeves (ow, this comment line here is really useless :laughing:).
To be clear: The language point was just an (unimportant) example, where `/msg` could be used for *politeness* rather than privacy. And I mentioned `/ignore` only to suggest another way to protect users from potentially distracting messages - no guideline changes implied. And hey, consistent decisions are not always unpopular.

ZeroKnight:
I'll continue working on this when a consensus has been reached on the expectation of these (private|direct|targeted) messages, and that the whole thing wont be dropped.

My personal stance has always been that in-game chat is not the place for extended and/or personal (to the point of privacy concerns) conversations because after all, you *did* launch a **game** and *not* IRC or an IM client. Whether this is implemented as "private" and not shown in demos, or implemented as a simple "whisper", targeted message, or whatever name makes you feel cozy does not matter to me. I understand both sides of the coin, but a consensus needs to be reached or the feature should just be scrapped.

qreeves:
I've already given my decision, call it whisper or targeted messaging or whatever. I'll take responsibility for the security so it isn't left to you to cop the shit.

ZeroKnight:
Alrighty, in that case I'll keep working on it then. Unless anyone has any objections I'll refer to these as "whispers", as it seems a somewhat popular name choice in other games I've played. Plus, semi-humorously speaking, even actual whispers can be heard by unintended ears :fearful:

In that case, I'll probably make the command `whisper`, but add a default `msg` alias for it, kinda like the `setmaster -> setpriv` alias.

ZeroKnight:
Thinking of ideas for how to format the "whisper" in chat... I've got a couple of thoughts.

* Similar to team messages (`<F00b4r (to team <icon>alpha)> lalalala`)

I could perhaps do: `<F00b4r (whispers)> lalalala` ("whispers" colored cyan or something). Fits in well with what is already implemented, gets the point across.

* Contextual icons. This goes beyond just the look of whispers, but regular/team chat as well.

Not too sure about this idea, but I came to it based off of @TheLastProject and @Jigoku's idea to prepend an icon instead of `(MSG)`. Since team messages use the team icon, it seems fine that whispers could use one as well. However, it would be the only message type to have a prepended icon, and throw off the general alignment of chat. So, why not enhance chat a bit and give each message type an icon?

**Regular Chat**
`[chat.png] <F00b4r> regular old chat lalalala`

**Team Chat** (*possible con is dropping the team name and just showing the icon to avoid misalignment*)
`[alpha.png] <F00b4r> team chat doodeedoo`

**Whisper**
`[whisper.png] <F00b4r> whisper whisper shhhh`

Could maybe even expand it to other events that happen in the console, like `alert.png` for team scrambling, something for map votes, team changes, etc...

Thoughts?

freezurbern:
@ZeroKnight That sounds like a good way of doing it. My only idea is for making sure that when it is sent to the log file, [whisper], [alpha], [omega], etc.. appear in the log instead of an icon.

ZeroKnight:
I'd like to have more people weigh in on the other style idea, but I've got [this](https://github.com/red-eclipse/base/pull/55/files#diff-255a3589e66b0c2c340777ff95a124b3R1041) going for now, similar style to team messages. Ex:

f00b4r sees:
`<f00b4r (whispers to [icon] ZeroKnight)> lalalala`

You see:
`<f00b4r (whispers to you)> lalalala`

ZeroKnight:
Okay, updated PR comment. Aside from any formatting suggestions and the alias, this is pretty much ready to be merged.

Ignore the failed semaphore build. It didn't seem to like when I force pushed my rebased commit to my fork.

qreeves:
Alright, let's give it a go.

PR #61, opened 2015-02-19
ignore command now tracks IP instead of CN Fix #56

### LINKED ISSUE #56: Use IP addresses for ignore
Currently, the client side ignore command works only on client numbers. Now that IP addresses are visible, it should use that instead to match against.

ZeroKnight:
Will now persist through disconnects, but like before all ignores are reset when the game is quit since they are not saved anywhere.

### DIFF
--- a/src/game/client.cpp
+++ b/src/game/client.cpp
@@ -845,24 +845,38 @@ namespace client
     ICOMMAND(0, clearlimits, "", (), addmsg(N_CLRCONTROL, "ri", ipinfo::LIMIT));
     ICOMMAND(0, clearexcepts, "", (), addmsg(N_CLRCONTROL, "ri", ipinfo::EXCEPT));
 
-    vector<int> ignores;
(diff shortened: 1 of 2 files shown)


ZeroKnight:
Would you prefer to have separate `ignore` and `ignoreip` commands, or have a single `ignore` command that can take either? 

Also, instead of relying on users modifying their autoexec we could write `ignore` commands into `config.cfg`, or even potentially create a new config like `ignore.cfg` and read/write ignores there with extra information like the player's name and hostname at the time. This could also provide the foundation for a new menu to better interface with ignores; like adding, deleting and listing them with aforementioned information.

qreeves:
I guess the specifics are up to you, if you're done for now I'll just pull this.

ZeroKnight:
For now yes. I'll see what I can cook up regarding the other things and submit another PR. Sound good?

PR #74, opened 2015-03-01
new map menu by @acerspyro

bonifarz:
As requested, I'll put this here for review as a pull request.
Please note that the two star icons by @acerspyro should be included for this, however I don't know if both, base and data can be addressed in a single pull request.
This is still work in progress, but with your feedback, this might turn into something useful :smile:

### DIFF
--- a/config/menus/maps.cfg
+++ b/config/menus/maps.cfg
@@ -1,617 +1,617 @@
-    // Made by Maxim (acerspyro) Therrien, completed on November 14th, 2014 (CC-BY-SA)
+// Made by Maxim (acerspyro) Therrien, completed on November 14th, 2014 (CC-BY-SA)
 
-    chosenmode = -1
-    chosenmuts = 0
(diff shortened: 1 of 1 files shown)


acerspyro:
Cool, thanks, I also have a new options menu coming up. So far, I've only designed, and am still coding in, the resolution menu.

bonifarz:
@acerspyro: That's nice. Maybe you will want to put your work in https://github.com/red-eclipse/community. I hope my modifications here serve as an inspiration for your menu. If you want to tweak the layout and replace this PR with one of your own, that's just fine for me. It is your creation, after all. I agree with the *postponed* label, as this will probably need some more testing, and we are all looking forward to the upcoming release!

bonifarz:
I guess it is better to close this request for now and keep the list clean.
This can be re-opened later, or resubmitted by the original author :wink: 
If I have another update for this menu, I'll put it in the community repo.

PR #75, opened 2015-03-02
add usage instructions(binds and missing gui*)

bonifarz:
NOTE: These points should be reviewed before merging.
* the scale and size of guiplayerpreview and guimodelpreview are not clear (scale is actually the image size, size does not seem to have an effect, not sure what is intended here.)
* for the mode arg of guieditor, I only mention 4, readonly. Is that enough for usage the instructions?
* is it correct to state the string arg of guieditor implies a textinit?
* is the comparison of guifill with guibackground okay? I guess the former should usually not be used.


qreeves:
* scale is the physical scale of the model that is rendered (as it would be if the player were in resize or the scale attribute of a mapmodel were altered).
* I have no idea/don't even remember anymore.
* Yeah pretty much, "str" is "initval".
* Each has their own use, but yes they function the exact same way.

Also, hope the new bigstring support is helping with the setdesc stuff.

bonifarz:
Okay, thanks.
Forgot about the new bigstring support, could be useful for descriptions of fields and sliders.
Will do a commit to the patch once I am done, such that the request can be merged&closed.
Maybe I just worried too much about details. Some useful descriptions are better than none.

PR #82, opened 2015-03-05
New NSIS Script

freezurbern:
Here's a new NSIS installer script, see the commit for full changelog. It should definitely be looked over and perhaps improved by someone else who knows how to use NSIS.


freezurbern:
I'm sure that MUI2 and other MUI things are actually doing anything.

PR #84, opened 2015-03-07
enable commands and aliases in the vars menu

bonifarz:
Use the vars menu to find commands and aliases, too.
For details, see the commit notes.

### DIFF
--- a/config/menus/main.cfg
+++ b/config/menus/main.cfg
@@ -39,27 +39,27 @@ newgui main [
         ] ] ]
     ]
     cases $guirolloveraction "showservers" [
-        guitooltip "displays the server browser for other online matches"
+        guitooltip "display the server browser for online matches"
     ] "showgui profile" [
-        guitooltip "edit your name, colour, and model type"
+        guitooltip "edit your name, colour, loadout and model type"
     ] "showgui edit" [
         guitooltip "access commands to use while editing a map"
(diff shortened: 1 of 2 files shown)


bonifarz:
I just realized that getvarinfo used for the vars menu can be applied to commands and aliases - that's cool! Usually if someone wants to search variables for a certain topic or keyword, aliases and commands are interesting to look up, too. With a few fixes to the menu, this seems to work pretty well. It does not make sense to "edit" aliases or commands in this menu, but hey, we can pass the search reult to the console :smiley: 
What do you think?

qreeves:
You should probably use `guibitfield` for the bitwise operators like `$varnotypes`, etc.

bonifarz:
Did so for testing, and left that as a comment, https://github.com/bonifarz/base/blob/patch-5/config/menus/vars.cfg#L11-36
So, the point is that two guibitfields (for varnotypes of commands and aliases) are better than one checkbox? Or should we add more options. Not sure how much space should be used for this.

qreeves:
If you're modifying a bitwise variable then you should always use the bit operators. This way the user can modify the value manually without losing their setting when pressing these buttons.

bonifarz:
Good point. And I guess it does not hurt to have checkboxes to narrow down the variable types. There seems to be enough room for that anyway.

PR #92, opened 2015-03-13
Translatable NSIS

freezurbern:
Read the commits, all the information is there. This separates custom strings into another file. Also removed license page (but added select language page).

### DIFF
--- a/src/install/win/lang.nsh
+++ b/src/install/win/lang.nsh
@@ -1,0 +1,35 @@
+﻿; Red Eclipse
+; Windows Installer Translations
+
+; Languages Supported:
+!insertmacro MUI_LANGUAGE "English" ;first language is the default language
+!insertmacro MUI_LANGUAGE "French"
+;!insertmacro MUI_LANGUAGE "German"
+;!insertmacro MUI_LANGUAGE "Spanish"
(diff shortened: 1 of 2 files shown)


qreeves:
My only concern is this might give users the wrong impression; they may believe setting the language will make Red Eclipse itself in their own language.

freezurbern:
That was my issue with translating the installer, but i forgot to mention
it. What's the point of having a translated installer if the game isnt in
any of the languages? Not to be insensitive and rude, if a user cannot read
the installers simple text, how will they understand the game. situation
kinda sucks.

Calinou:
The inverse is worse – where the installer lacks language but the game has languages (this is the case for GIMP with me; no French installer, but French program is here).

Evropi:
@Calinou, "the inverse is worse" also implies that you don't like this situation either -- of a translated installed but untranslated program. Do you personally believe it's a positive or a negative? I wouldn't expect the game itself to be made translatable any time soon.

freezurbern:
Perhaps include a notice in the installer that only shows for non English
that says the game content is in English? I also cant see the game being
translated anytime soon. All those pictures!

freezurbern:
I made another PR #93 to get the license page removed seperately. Perhaps we can get 1.5 released soon and work towards translations for 1.6

PR #103, opened 2015-03-22
Fix loadout menu for tips and key binding menu

zaquest:
The problem this pull request solves is that keys that are bind in gui to open loadout menu dont work and dont show up in tips.


shirepirate:
already fixed:
https://github.com/red-eclipse/base/pull/102

shirepirate:
thank you

freezurbern:
Thanks for catching those things I didn't!

PR #108, opened 2015-03-23
NSIS updates after 1.5.0

freezurbern:
I heard more than a few people giving feedback that the Windows installer would pause for an unusually long time after verifying itself, but before opening the first page. I determined this was caused by the /SOLID flag being used for compression of the game files. This PR makes the installer open very quickly (less than half a second as opposed to over twenty)
Also, I have included a much prettier finish.bmp created by @acerspyro. This looks a lot more professional than my attempt at it.

Note: These settings only increased the installer's size by 3 megabytes.


shirepirate:
kinda too late for this to be incorporated into the 1.5 release, pity.

freezurbern:
The reason this PR exists is because testing was finally done widespread by
the release :) unfortunate but hey its ready for anything we do next

qreeves:
We'll push out 1.5.1 on the 29th I guess.

PR #115, opened 2015-03-24
Fix the OS X compilation for Red Eclipse 1.5

kerio92:


src/game/game.h
```diff
@@ -469,7 +469,7 @@ struct verinfo
     template <class T>
     void get(T &p)
     {
-        string text = "";
+        string text;
```
qreeves: Don't do this, you're leaving the field uninitialised. This is NOT std::string (see `src/shared/tools.h`).

src/xcode/SDLMain.m
```diff
@@ -193,22 +197,22 @@ static void setupWindowMenu(void)
 /* Replacement for NSApplicationMain */
 static void CustomApplicationMain (int argc, char **argv)
 {
-    NSAutoreleasePool   *pool = [[NSAutoreleasePool alloc] init];
-    SDLMain             *sdlMain;
+    NSAutoreleasePool	*pool = [[NSAutoreleasePool alloc] init];
+    SDLMain				*sdlMain;
```
qreeves: Was it really necessary to add all this whitespace, if anything I'd rather it be reduced to one space, not expanded.

bin/redeclipse.app/Contents/MacOS/redeclipse
```diff
@@ -1,15 +0,0 @@
...
-APP_OSXVER=`sw_vers -productVersion | sed -e 's/\.//g'`
-if [ ${APP_OSXVER} -lt 10000 ]; then
-    let "APP_OSXVER*=10"
-fi
-
-if [ ${APP_OSXVER} -ge 10600 ]; then
-	exec "${0}_universal" "$@"
-else
-	exec "${0}_compatible" "$@"
-fi
```
qreeves: You say on IRC you have no idea how to make a ".app", yet you go messing around with these files by deleting them. We distribute the SDK files because otherwise people have to manually retrieve them (and we don't install any system-wide).

By deleting this shell script you also remove any possibility that this will ever run on anything lower than OS X 10.6. In fact, that's the whole problem here. While it is easy to make everything run on the latest OS X, there are a lot of people who don't pay the eye-gouging prices to upgrade their OS for marginal gains.

src/xcode/SDLMain.m
kerio92: As i wrote in the commit, this is the latest SDLMain.m from SDL 1.2.15, i didn't touch it in any way

qreeves: Heh sorry, didn't notice it was the SDL headers

qreeves:
On reflection, 10.6+ support us better than no support. Can you verify it compiles the "Release" build?

src/game/game.h
kerio92: I'm not entirely sure why clang chokes on it, apparently "" is too long of an initializer. I'm changing that to `string text = {0};` I guess.

kerio92:
I can only check if it builds and runs on my machine, though (10.10 on x86_64).

PR #117, opened 2015-03-25
FreeBSD fixes

AMDmi3:


src/Makefile
```diff
@@ -3,10 +3,10 @@ APPCLIENT=$(APPNAME)
 APPSERVER=$(APPNAME)_server
 
 #CXXFLAGS= -ggdb3
-CXXFLAGS= -O3 -fomit-frame-pointer -ffast-math
+CXXFLAGS?= -O3 -fomit-frame-pointer -ffast-math
```
qreeves: What does this do?

src/enet/Makefile
```diff
@@ -1,4 +1,4 @@
-CFLAGS=-O3 -fomit-frame-pointer
+CFLAGS?=-O3 -fomit-frame-pointer
```
qreeves: We don't maintain "eNet", you'll have to prod @lsalzman about this one.

src/Makefile
AMDmi3: CFLAGS/CXXFLAGS may be set by user or build environment and should not be overridden in that case. Default value is OK though. This change preserves default value, but respects user setting if there is one.

src/enet/Makefile
AMDmi3: This Makefile is probably maintained by RE contributors, as it doesn't exist in enet repo.

qreeves: Actually, this file shouldn't be in the repo at all, it is generated by './configure', haha.

AMDmi3: There's no longer configure in the repo, no? The looks to be handwritten.

PR #118, opened 2015-03-25
fix guieditor not rendering text if 1st line empty

### LINKED ISSUE #78: guieditor issue when first line is empty
Just a small detail I noticed when experimenting with guieditor:
If the first line of the loaded file is empty, no text is shown in the editor field.
However, the field has the correct scrollbar and moving around will also update $textcurrentline correctly.

ZeroKnight:
Closes #78


qreeves:
I think this fix might actually cause issues elsewhere. There can be multiple lines, all blank. Will take a better look at this PR when I fix my laptop.

PR #124, opened 2015-03-26
zoomstyle variable to allow for multiple styles of zooming, currently 1.4 style and 1.5 style.

shacknetisp:
This is due to the discussion at http://redeclipse.net/forum/viewtopic.php?f=4&t=640


qreeves:
A better way to do this would be to do the same thing I did when I added the local flag to user privileges (which is what I was going to do). If you want to try again using that model, that'd be great.

shacknetisp:
I'm not sure what you mean, would that be changing the zoomstyle variable to a bit field or something entirely different?

qreeves:
See how 'gamestate::privilege' is used. If you don't get it, just create an issue and I'll do it when I get a chance (and am ready to break protocol).

PR #138, opened 2015-04-03
enhance loadout menu

bonifarz:
on http://redeclipse.net/forum/viewtopic.php?f=4&t=647 @shirepirate wrote:
>  the way loadout is handled currently, it's not very intuitive, rather confusing and even at times frustrating.
> [...] after changing a loadout in-game, players are left with little or no explanation that it won't take effect until they respawn

* notify players about respawn for weapon change if their state is *alive*
* bring back the option to toggle display of all weapon slots
* show weapon names if extra slots are hidden
* keep a nice layout (with another framed box)

### DIFF
--- a/config/menus/profile.cfg
+++ b/config/menus/profile.cfg
@@ -1,12 +1,11 @@
 profilebuttons = [
-    guicenter [ guifont "emphasis" [
+    guilist [ guifont "emphasis" [
+        guistrut 2
         guibutton "^fgok" [
             setinfo $playerprevname $playerprevcolour $playerprevmodel $playerprevvanity $playerprevlweap
             cleargui 1
         ] []
-        guistrut 0.5
-        guispring 1
-        guistrut 0.5
+        guistrut 3
         if (needname) [
             guibutton "^fomain menu" [showgui main]
         ] [
@@ -16,7 +15,7 @@ profilebuttons = [
 ]
 
 profilepreview = [
-    guiplayerpreview $playerprevmodel $playerprevcolour $playerprevteam $playerprevweap $playerprevvanity [playerprevinherit = 1; showgui playerprev] 7.5 1 1
+    guiplayerpreview $playerprevmodel $playerprevcolour $playerprevteam $playerprevweap $playerprevvanity [playerprevinherit = 1; showgui playerprev] 7.6 1 1
 ]
 
 playerprevinherit = 0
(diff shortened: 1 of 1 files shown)


bonifarz:
loadout138

Evropi:
A good start. The text should be right next to the weapon icons column to better indicate the relationship between the two.

bonifarz:
Good point. Though, if I do not center the text, there is quite a bit of empty space to the right, as I keep the size of the frame constant (which actually looks rather complicated, with some blank images and spacings). Maybe I can add some spacing to the right of the box accordingly, if you think that is better.

bonifarz:
Do you think this works better? Note: I slightly altered the OK/cancel buttons, too.

shirepirate:
let's try something closer to this instead, as the larger weapon images are a bit more attractive:
http://redeclipse.net/forum/download/file.php?id=990&t=1&sid=0df2ee75cbceabfa554ea9e21ddc61d5

create sort/ordering that extends upward above these images, either radio buttons or checkboxes or even a custom gui image. then it would be no problem to reintroduce a column for random again, but wouldn't be as overwhelming as the current configuration.

"show all weapon slots" can remain as in this PR, but probably the unavailable slots should be signified in some way when they are in fact not available (i.e., the available slots could have a white number, the unavailable slots could have a darker shaded number)

"save loadout" would be restored so that the player would have to check the option in order to not be asked again in the future (rather than simply clicking OK in profile)

if we can convince @qreeves to come up with a different solution to prevent network abuse, it won't be necessary to include the weapon respawn warning, you would simply respawn automatically as before

bonifarz:
Yeah, I intended to get back soon to this experiment of sorting a list of loadout weapons. I think it can be more intuitive, and might even work without modifying the sources (setprofile). Checkboxes above or below the images seems just as confusing as the old style grid, though :wink: 
I'll commit another draft here later, and then we can discuss how to iterate the enhancements.
In the main menu, we might simply reword *profile* to *profile and weapons*.

bonifarz:
loadout experimental

qreeves:
Is this ready for merging? If so, let's do it.

bonifarz:
Not the latest commit here:
> Experimental: Although two random slots are shown here, one of them uses weapon id 1, which will actually be ignored and results in use of the next weapon in the list. So, this is to test and discuss a different idea for the loadout buttons in the profile menu.

So, I put this here to see if the buttons are useful at all like this (click -> move to 1st slot, left click -> move to last slot). If we want to do it this way with two random buttons, we have to make sure how to deal with them (using id 0 and 1 or something else).
Side note: The previous commit (0401c0e *show all weapon slots*) was ready to merge, but I guess it is not such a huge improvement (rather a step towards the old style).

qreeves:
You should probably recreate this PR, it's too much of a pain in the ass trying to partially merge it.

bonifarz:
I guess the experiment to simplify the loadout buttons is not that interesting for now, so it can be kept in this closed PR here.

PR #140, opened 2015-04-03
Bot command quick-chat menu

JeffBobbo:
Bot command quick-chat menu, opened via the B key.
Automatically disabled for demo, edit and race.
Attack and defend options appear in same places regardless of game mode.
Sends the commands to team chat (so those pesky enemies don't know what you're planning)

### DIFF
--- a/config/compass.cfg
+++ b/config/compass.cfg
@@ -2,6 +2,7 @@ newcompass main "textures/menu" [
     compass "menus" [showcompass menus]
     compass "voice" [showcompass voice]
     compass "team" [showcompass team]
+    compass "bot" [showcompass bot]
     compass "clear" [clearcompass]
 ]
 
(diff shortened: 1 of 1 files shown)


qreeves:
Probably better to just check if the team mutator is selected.

bonifarz:
Or check if any bots are currently in your team. Also, I guess it would be cool to send commands not only to all bots, but also have an option to prefix a bot name. In return, this may be combined with a compass for the whisper command. There's many options to deal with this. Let us figure out what works best.

JeffBobbo:
@queeves I dunno how to check for the team mutator, if you let me know I'll change it.

@bonifarz I did think about this (but actually for the "defend <player>" option). The compass menu only supports 8 items (the ninth appears in the middle, never tested with ten). I get the feeling adding the ability to prefix not names to these commands would be cumbersome for the user to use.
And as always, what to do if there is more than 8 bots/players? Continuation compass on the next 'page'?

bonifarz:
If there are more than 8 entries, they get listed below, but that's not really the point of a compass, and the bigger difficulty is to assign reasonable and unique hotkeys. Right, one solution is to use the same key to open another compass ("next page"). But I agree, if it gets too complex, it kills the purpose of a quick command compass.

qreeves:
`if (& (mutators) $mutsbitteam)`

bonifarz:
I think $mutsbitteam is deprecated, but you can still check for (implied) *ffa*, right?

    if (! (& (mutators) $mutsbitffa))
Just in case, here's a snippet to get the bot names of your team:

    mybots = ""
    looplist cn $listclients [
        if (& (isai $cn) (= (getplayerteam 1) (getclientteam $cn))) [
            append mybots (getclientname $cn)
        ]
    ]

qreeves:
Oh yeah, @bonifarz is right :P

qreeves:
Merging this now so people can test it.

PR #142, opened 2015-04-03
Update usage.cfg

IceflowRE:
Add
- mastermode
- masterlock

Update
- Everything which was example: classic bath and not classic mode
- sort everything from A to Z (loops are still at least and the material comes after air)


IceflowRE:
Whatever idk what this material loop should do? Isnt it the normal setdesc? Why the loop?

bonifarz:
Do you have a copy before you sorted the descriptions? Sorting is trivial, and I guess the ordering and grouping of commands was there for a reason.
Material loop: Because we have different commands /water, /water2, /water3 etc.
I agree a good description of *mastermode* is useful, unless we want to change (rename or split) this command.

IceflowRE:
I cannot see a really linear sort by groups or something, yeah it is a little bit sorted but not all.

PR #148, opened 2015-04-08
update default texture scale (0,5)

IceflowRE:
because the size was doubled


Calinou:
Good idea to have an higher-resolution default texture.

I would also try making it brighter.

PR #158, opened 2015-04-18
improve display of votes count

bonifarz:
I realize that with  #156, the votes count in the guitab header and in the box at the top right are the same.
So, what about showing the total of votes (i.e. the player count) in the header, 
and keeping the number of *games suggested* in the box?


qreeves:
Worth a shot I guess.

PR #160, opened 2015-04-20
A getdemo name parameter, getdemo and listdemos entries in usage.cfg

shacknetisp:
Breaks Protocol


qreeves:
Cube 2 code does not use free/strdup. Please use delete/DELETE and newstring.

I'm not sure if a vectored list like this is a good idea either, everything is rather asynchronous and you can't be guaranteed that it is responding to the correct reply. I'd say the protocol needs a slight modification to include a unique request ID when getting the demo from the server.

shacknetisp:
I deleted the wrong branch by accident.

src/game/game.h
```diff
@@ -4,7 +4,7 @@
 #include "engine.h"
 
 #define VERSION_GAMEID "fps"
-#define VERSION_GAME 226
```
qreeves: I'd rather pull requests left the version information alone so I can do it without getting conflicts, can I get you to remove this part please?

qreeves:
Please rebase this branch on master so I can merge.

PR #163, opened 2015-04-22
demoautoserversave variable for the server to auto save demos, closes #162

### LINKED ISSUE #162: Save demos to disk on server
From: http://redeclipse.net/forum/viewtopic.php?f=4&t=145&p=1166

> I would really appreciate the option of saving demos to disk for the server, knowing full well that it could result in a lot of disk usage over time.
I think the only issue with not implementing this would be the effort it would take to code it. 
>-freezurbern

---

> Demo's are only stored in the server's memory, the idea to save them to disk is a valid request. 
-qreeves

---

More thought on this, perhaps a variable called `demosave` or `demodisk {0/1} // Save demos to disk on the server. NOTE: This can take up considerable disk space.` Perhaps the variable for limiting demos that can be retrieved could be modified to either how many are saved in memory and how many are saved to disk.

shacknetisp:
Enable **demoautoserversave** to automatically save demos to `<serverhome>/demos`.
Limiting the space used by this is done with **demoserverkeeptime**, which (if greater than 0), will delete all auto-saved demos older than **demoserverkeeptime** hours old.

closes #162

### DIFF
--- a/config/usage.cfg
+++ b/config/usage.cfg
@@ -22,6 +22,7 @@ setdesc "demolock" "determines who may record demos;^n0 = off, 1 = player, 2 = s
 setdesc "democount" "determines the maximum amount of demo files that may be saved simultaneously on the server (deletes old demos if exceeded)" "<value>"
 setdesc "demomaxsize" "determines the maximum size of individual demo files that may be saved on the server" "<kilobytes>"
 setdesc "demoautorec" "determines if demos are automatically recorded each match" "<value>"
+setdesc "demoautosave" "determines if the server automatically saves demos to disk" "<value>"
 setdesc "mastermode" "set a specific mode to the server;^n0 = open (everybody can join), 1 = veto (vetolock determines who can force a vote), 2 = locked (only moderator or higher can spawn), 3 = private (only global privs and players with the password can join the server)" "<value>"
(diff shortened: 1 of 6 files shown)


shirepirate:
is this variable supposed to be for the benefit of clients?

without documentation it's hard to know, but would the demoinfo command be more useful for the filenames?

shacknetisp:
The variable is for servers to keep a permanent store of demos on disk.
demoinfo is a client command.

shirepirate:
ah, i was under the impression files were already created

shirepirate:
is there a var for a client to automatically save available demos?
might be another interesting var

qreeves:
I'd feel a whole lot better if there were some kind of limit placed on the number of files this could potentially generate. This could use up a lot of space. Is there a reason to have the server write them anyway? Due to the volume, it's not something you could review all of. Players in the game are better able to determine if it is "save-worthy" or not.

shacknetisp:
The benefit of this would be that you could still access demos if the server restarts or crashes.
The space problem could be solved by only keeping the last _n_ demos or deleting demos older than some specified time.

freezurbern:
Can we get documentation on `demoserverkeeptime` in usage.cfg?

bonifarz:
@freezurbern: merged via #195 by @shacknetisp 
>if greater than 0, only keep auto-recorded demos younger than demoserverkeeptime seconds

PR #164, opened 2015-04-22
New mapinterface based on @acerspyro

IceflowRE:
Add
- show vote menu only online
- tooltip who can still suggest maps who arent in the rotation
- optical changes in the vote menu (percent,, how much has voted of all
players)
- tooltip which shows the mutators

some more if i forgot something

Thanks for really much help @shacknetisp (Beha) and @Bonifarz

PR: WHat do you say about it?

### DIFF
--- a/config/menus/maps.cfg
+++ b/config/menus/maps.cfg
@@ -176,15 +176,16 @@ mapselection = [
 				guistrut 0.2
 				guimerge 35 [
 					guibackground 0x444444
-					guiimage "textures/info" [] 1.5 1
+					guinohitfx [
+						guiimage "textures/info" [] 1.5 1
+					]
 					guistrut 1
 					guilist [
 						guistrut 0.5
 						guitext "no matching maps"
 						guifont "little" [ guitext "^fdtry another search filter"]
 					]
-					guispring
-					guiimage "textures/guiexit" [ mapsearch = "" ; mapsearchon = 0; fillmaparray] 0.5
+					guispring 1
 				]
 			]
 			if (< $j 8) [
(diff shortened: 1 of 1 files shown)


IceflowRE:
Oh forgot they need the staron and off images in data/textures/
star
staron
Images were created by @acerspyro i think

acerspyro:
Yes, I made the icons.

IceflowRE:
Ive found something which should be fixed before merging, i will write a comment if ihave done it.

IceflowRE:
Done it!! (After long long time of thinking and working D: )

shirepirate:
it's a lot to scroll through without much added benefit.
it takes longer to scroll through them and you can't see the map images very well.
for the inconvenience there isn't a lot of added benefit.

instead, an alternate view of the maps menu should be implemented that uses larger scaled images for a nice change in visuals (one that does not replace the current text-only menu, but provides a visual option instead).

in this visual style menu it's not necessary to have the preview box anymore, as maps should be visible enough at that scale. also the applicable modes/muts for that map should be displayed as a tooltip instead.

the random/reset/copy buttons are too small. what is the function of the copy?

is the playersonserv alias a necessary thing to have? what is it's purpose and application?

the "server type" could perhaps be clarified a little better, to make it easier to understand. something like "offline play," "start private server" & "start public server"

additional textures should be included in the pull request. it would be nice if they were a slightly higher res as well.

closing this, create another pull request in the future if you make changes.

bonifarz:
> an alternate view of the maps menu
Ah, you mean that grid of preview tiles, similar to the textures menu? I could have another look at it. As far as I remember, the disadvantage was the "loading" time on weaker hardware, as it takes a moment to scale the images.
I still have not compared what exactly icy has added to the previous PR #74  I derived from acerspyro's earlier work, so I cannot comment on this for now.
Concerning the inclusion of the icons, I don't have a clue how this can be done in a single pull request, as there are two distinct repos, base and data.

IceflowRE:
>it's a lot to scroll through and it doesn't seem to be worth it.
>it takes longer to scroll through them and you can't see the map images very well.
>for the inconvenience there isn't a lot of added benefit.

the small map images arent for looking the detail, they should only help to remember better which map is behind this name (and that you dont have to click everytime to show the image)

>in this visual style menu it's not necessary to have the preview box anymore, as maps should be visible enough at that scale. also the applicable modes/muts for that map should be displayed as a tooltip instead.

To many tooltips gets annoying, you have to look everytime that your mouse isnt on a thing which opens a tooltip and blocks the view. A higher image scale has the consequence to have fewer maps in the list or if you chose a grid, you cannot implement the names good.

>the random/reset/copy buttons are too small. what is the function of the copy?

They shouldnt really big, because when do you use them? not really often (ok reset more often)
It copies the current play/ed mode/ muts.

>is the playersonserv alias a necessary thing to have? what is it's purpose and application?

playersonserv returns the numbers of players of the server.

>the "server type" could perhaps be clarified a little better, to make it easier to understand. something like "offline play," "start private server" & "start public server"

I think these buttons can be removed...

PR #165, opened 2015-04-23
fix the server browser for systems without gethostbyname_r()

jonathangray:
enet_address_set_host() is only thread safe on systems with
gethostbyname_r().  Don't unlock the resolver mutex when calling
it to prevent crashes in the server browser seen on OpenBSD.


qreeves:
I'm not familiar enough with this to make the decision.

lsalzman:
Should be resolved now by using getaddrinfo internally, which is advertised as thread-safe on OpenBSD.

qreeves:
Thanks :smile:

PR #167, opened 2015-04-24
Fix bitmap iterators signedness

ienorand:
Change x & y iterator signedness to match the unsigned ints (rows and
width of the bitmap) which they are being compared with.

Fixes [-Wsign-compare] compiler warning.
---
Just to make sure this goes through some review I'm doing it as a pull request.

### DIFF
--- a/src/shared/cube2font.c
+++ b/src/shared/cube2font.c
@@ -275,7 +275,7 @@ void writetexs(const char *name, struct fontchar *chars, int numchars, int numte
         for(i = 0; i < numchars; i++)
         {
             struct fontchar *c = &chars[i];
-            int x, y;
+            uint x, y;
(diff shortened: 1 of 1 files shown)


src/shared/cube2font.c
```diff
@@ -275,7 +275,7 @@ void writetexs(const char *name, struct fontchar *chars, int numchars, int numte
         for(i = 0; i < numchars; i++)
         {
             struct fontchar *c = &chars[i];
-            int x, y;
+            unsigned int x, y;
```
qreeves: We use the 'uint' convention in C2 sources.

ienorand:
Amended to use 'uint' over 'unsigned int'.

PR #172, opened 2015-05-02
demoautoclientsave variable, runs getdemo after every game if set

shacknetisp:
Breaks protocol.


qreeves:
If you're going to make a new network command, you may as well replace the `srvoutf` by putting the info into the network packet instead.

shacknetisp:
That would stop the message being written to the server output/IRC though.

qreeves:
You'd replace it with a `relayf` in that case.

qreeves:
Same deal, please rebase on master for merging.

PR #178, opened 2015-05-08
A hidden variable for servers.

shacknetisp:
The variable is transmitted via `addserver`.
Clients will only see servers with 'hidden=1' if they enable `listhidden`.
The purpose of this is to allow servers to be modified heavily and still be listed on the master server, while giving new players the best experience.

Idea from http://forum.freegamedev.net/viewtopic.php?f=53&t=2998

### DIFF
--- a/src/engine/engine.h
+++ b/src/engine/engine.h
@@ -462,7 +462,7 @@ extern void localconnect(bool force = true);
 extern void localdisconnect();
 
 // serverbrowser
-extern void addserver(const char *name, int port, int priority = 0, const char *desc = NULL);
+extern void addserver(const char *name, int port, int hidden = 0, int priority = 0, const char *desc = NULL);
 
 // client
 extern char *connectname;
--- a/src/engine/master.cpp
+++ b/src/engine/master.cpp
@@ -50,7 +50,7 @@ struct masterclient
     int inputpos, outputpos, port, numpings, lastcontrol, version;
     enet_uint32 lastping, lastpong, lastactivity;
     vector<authreq> authreqs;
-    bool isserver, isquick, ishttp, listserver, shouldping, shouldpurge;
+    bool isserver, isquick, ishttp, listserver, shouldping, shouldpurge, hidden;
 
     masterclient() : inputpos(0), outputpos(0), port(MASTER_PORT), numpings(0), lastcontrol(-1), version(0), lastping(0), lastpong(0), lastactivity(0), isserver(false), isquick(false), ishttp(false), listserver(false), shouldping(false), shouldpurge(false) {}
 };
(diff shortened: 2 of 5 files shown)


Calinou:
:+1:

nyuszika7h:
Sounds good.

shirepirate:
as i see it, the guidelines for connecting to the master server are the guidelines for connecting to the master server. they wouldn't change with this implemented, so what would be the point? wouldn't it be better for servers to be more visible?

shacknetisp:
It would give servers with unauthorized modifications, (such as the one mentioned [here](http://redeclipse.net/forum/viewtopic.php?f=4&t=324)) the ability to be listed. This would, however, require an amendment to the guidelines to allow for hidden servers to use unauthorized modifications.

From the [Old Trac Open Issues](http://redeclipse.net/forum/viewtopic.php?f=9&t=8) topic:

>**Master Server Policy Follow-ups**
 >*The problem*:
 >http://forum.freegamedev.net/viewtopic.php?f=53&t=2995
 >A vocal minority has problems with the level of moderation done by those with auth keys, some also  disagree with the guidelines set out by the project regarding the expectations of those using the master server.
 >*The suggested solutions*:
 >http://forum.freegamedev.net/viewtopic.php?f=53&t=2998
 >Allow server owners to apply certain flags to their configuration, defining them as "unmoderated"/"modified"/etc and restricting their view-ability on the master server list.

shirepirate:
i'm not voting for easier viewability of servers with terrible variables like that one. but aside from that, hidden or not, they would still be listed and "unauthorized modifications" is far too vague for that to slide. the vast majority of unauthorized modifications would be those that couldn't be allowed in any case regardless, not simply stupid weapon variables like the example.

it's easy enough to set up a private server in any way you want and share it with friends / publish it on a website. and if you want it publicly listed start your own master server. a modification of the server menu to make it easier to enter in custom master servers and switch between them sounds entirely acceptable, this would require considerable effort by someone who would wish to play on such servers as opposed to a variable which one could accidentally stumble across.

qreeves:
Yeah, it was never my intention to just let all crazy modifications on the master server. I've never made any progress on this issue because I don't feel it to be a high priority (for the project). We're focused on making *this* game, not helping others write their own and possibly giving users a bad impression of the game. This is why we defend the master server so rigorously with set guidelines; *quality assurance*.

In my opinion, any modifications that would need special permission to be on the master server would be better off as a content proposal for the game itself. I'd rather advocate open source than have to deal with each and every modified server with unshared source code.

shacknetisp:
I probably should have made this a poll on the forum first, I see why it is not wanted.

Issue #175, opened 2015-05-06
Install method for Linux


qreeves:
Currently, we have a few inexperienced Linux users (mainly Ubuntu) who have been tripped up by the fact that online support has ended for the version distributed by their respective distributions. These users typically don't know how to use the command line, and some are even confused why _our_ packages don't replace the shortcuts from the ones provided by their distribution.

What we could use is an _install.sh_ or something similar, which can set up a _$HOME_ hosted install with shortcuts the user can use. This would be separate from the makefile as this is an _install method_ done from the package we distribute.

Jigoku:
As for shortcuts and menu items. You'd want to generate a desktop file named `red-eclipse.desktop` which would be linked to `~/.local/share/applications` and `~/Desktop`.

Something like this would make shortcuts:
```
#!/bin/sh
cat <<EOF > red-eclipse.desktop
[Desktop Entry]
Name=Red Eclipse
Comment=Aurora Edition (1.5)
Exec=${PWD}/redeclipse.sh
Icon=${PWD}/data/textures/emblem.png
Path=${PWD}redeclipse-1.5/
StartupNotify=true
Type=Application
EOF

ln -s ${PWD}/red-eclipse.desktop ~/.local/share/applications/red-eclipse.desktop
ln -s ${PWD}/red-eclipse.desktop ~/Desktop/red-eclipse.desktop
```
If you run this script from within the location you extracted the red eclipse tarball (ie; install.sh), it makes the local install launchable via the menus/desktop search tools, and also as a desktop icon.

If the user were to delete the local install, the shortcuts also disappear due to being symlinks.

acz13:
and symlink redeclipse.sh into /usr/bin? maybe?

`ln -s ./redeclipse.sh /usr/bin/redeclipse.sh`

Jigoku:
It seems *.desktop files don't play nice with symlinks, and i noticed a pre-existing redeclipse.desktop.am that includes translations in the src/ directory, using that as a base, the scirpt in the pull request simply adjusts the paths for the local folder install.

ienorand:
I've done a quick test with one method to do it in ienorand/base@57862f1 where it's just installing a desktop file which links back to the unpacked tarball.

I don't feel like there's any particularly good solution to this issue to be found...

TheLastProject:
What is wrong with .deb and so packages? Those cause a pretty user-friendly installer screen when opened (screenshot on KDE as example):
[...]

PR #181, opened 2015-05-12
script to create desktop shortcuts for home folder install (*nix)

Jigoku:
This should work fine. Using the existing `src/install/nix/redeclipse.desktop.am` simply replaces the @APPNAME@ variables for the location of the local install.

It creates a `redeclipse-local.desktop` in both `~/.local/share/applications` and `~/Desktop` so the local install can be launched via a shortcut in the users application menu or from the desktop.
As discussed here: https://github.com/red-eclipse/base/issues/175

### DIFF
--- a/src/game/server.cpp
+++ b/src/game/server.cpp
@@ -4547,11 +4547,11 @@ namespace server
     }
 
     int lastquerysort = 0;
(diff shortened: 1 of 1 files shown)


Jigoku:
oops, closed by accident

PR #197, opened 2015-06-12
Residual changes as discussed/voted on forum

JeffBobbo:
http://redeclipse.net/forum/viewtopic.php?f=4&t=707


qreeves:
@shirepirate and I weren't really satisfied with the results of the poll and thought it best to leave the residuals alone.

PR #213, opened 2015-06-27
Improved weapon loadout menu & weapon names

JeffBobbo:
Moved <weapon>longname to <weapon>name.
Improved the <weapon>names.
Removed some shoutiness in profile settings (block caps).
Added clarifcation line about what the loadout is in loadout setting menu.
Added <weapon>desc[12] providing a short description of a weapon, including tooltip in the loadout menu.

some weird newline stuff that slipped in.

### DIFF
--- a/config/menus/profile.cfg
+++ b/config/menus/profile.cfg
@@ -124,11 +124,12 @@ newgui profile [
             guistrut 3
             guilist [
                 guistrut 0.75
-                guifont "emphasis" [ 
-                    guicenter [ guitext "CHOOSE YOUR WEAPONS" ]
+                guifont "emphasis" [
+                    guicenter [ guitext "Choose your weapons" ]
                 ]
                 guistrut 0.5
                 guifont "little" [
+                    guicenter [ guitext "you will spawn with these weapons" ]
                     guicenter [ guitext (format "you can carry ^fs^fc%1^fS %2" $maxcarry (? (= $maxcarry 1) "weapon" "weapons")) ]
                 ]
                 guistrut 0.5
@@ -157,7 +158,7 @@ newgui profile [
             ]
             guibar
             guistrut 2
-            guilist [ 
+            guilist [
                 guibackground
                 guicenter [
                     guistayopen [
@@ -177,7 +178,7 @@ newgui profile [
                                     guistrut 0.125
                                     guiimage [textures/question] [loadweaps @w2 0] 0.75 0 [textures/blank] [] (? (= $w3 0) 0xFFFFFF 0x808080)
                                     guistrut 0.125
-                                ] 
+                                ]
                                 loop w1 $weapidxloadout [
                                     w4 = (+ $w1 $weapidxoffset)
                                     w5 = (at $weapname $w4)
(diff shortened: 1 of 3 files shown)


shirepirate:
the short names exist for a reason, that is why the long names are held in another variable as requested by @qreeves 

we're both pretty satisfied with the names at this point, anyway these names don't seem to comply with your primary complaint of a name describing only 1 function of the weapon (a rocket isn't normally guided, a grenade doesn't always stick, etc). we'd discussed laser rifle and might change it to this at some point, but leave them for now.

the descriptions are ok if you wish to create a separate pull request.

the formatting of capital letters is uniform with many of the other menus and meant to stand out immediately, don't change it.

JeffBobbo:
>the short names exist for a reason, that is why the long names are held in another variable as requested by @qreeves
@shirepirate could you explain what this reason is then? This is the kind of thing that should be in comments or documented somewhere so people don't make this mistake.

As for
>your primary complaint of a name describing only 1 function of the weapon
You plainly ignored my suggestion to just name rocket and grenade "rocket" and "grenade". So clearly I had to come up with something, others thought the suggestion of "frag grenade" was too CoD-like.

As for the reasons for changing the weapon names, here's some I've come up with:
"flamethrower" is a single word,
"electric zapper" sounds better than stun zapper, stun is also used on mine
"laser rifle", as discussed on IRC is a more fitting name
"thermite grenade" doesn't fit at all (thermite doesn't explode), although I've noticed newer players don't realise grenade2 is sticky, this should help with that.
"guided rocket", again with grenade2, some players (even those that have been playing for months) don't realise rocket2 is guided.

acerspyro:
Shire, tell me, you seem to be hating everything Bobbo does. If someone else proposed the exact same thing, you would probably accept it, or change one thing and accept it. Care to share why?

shirepirate:
the descriptions you've added will help in that understanding, no need to change the names for that reason. quin changed the longname flame thrower to be in uniform with the other names as multiple words and not stand out as different.

the name variables define every other weapon variable and so having them be needlessly long would be tedious i'd imagine, there may be other game-breaking reasons but you'd have to ask qreeves.

>Shire, tell me, you seem to be hating everything Bobbo does. If someone else proposed the exact same thing, you would probably accept it, or change one thing and accept it. Care to share why?

my opinion of bobbo has no bearing on whether i'd accept this commit, though his reputation for opening false issues does proceed him so i do proceed with caution when his name comes up. i'd give all the same reasons for this PR needing change as i've just layed out if you had submitted it, or someone else. please don't speculate.

JeffBobbo:
>the descriptions you've added will help in that understanding, no need to change the names for that reason. quin changed the longname flame thrower to be in uniform with the other names as multiple words and not stand out as different.

There's no way to see the rocket description yet, I added this for future expansion of some gui about the weapons.

>the name variables define every other weapon variable and so having them be needlessly long would be tedious i'd imagine, there may be other game-breaking reasons but you'd have to ask qreeves.

This isn't true, I actually tested this as I suspected it might mess something up, but it turns out to be completely fine, you can change this variable to whatever you want, it's only used for display purposes. Although I must confess I don't know where the game gets the names for the variables from, it's not from <weapon>name, which would make sense considering to use <weapon>name you need to know what <weapon>name is.

shirepirate:
>Although I must confess I don't know where the game gets the names for the variables from
probably from /weapname, but regardless if it would break other features or not (quin implied it would) having short names may prove to be useful, so leave them.

JeffBobbo:
>probably from /weapname, but regardless if it would break other features or not (quin implied it would) having short names may prove to be useful, so leave them.

I tested that too, doesn't make a difference. And weapname provides us with a name to get a shortname anyway, as they're the same.

JeffBobbo:
@shirepirate Since you haven't answered all the queries, could this be re-opened so @qreeves could expand on what you haven't please?

qreeves:
I specifically asked for *name* to be kept as is, for any instances where the *longname* might be too long (I can only really determine this by seeing what breaks). I also added the condition that the long name should be two words for consistency (RE: *flamethrower*).

The fact of the matter is that this is shirepirate's topic, he came up with the idea and cleared the specifics with me. I gave him some conditions, and he has been working according to them. Not everything requires a million opinions as soon as it is created (*too many cooks..*), and I would appreciate if we could fiddle with them a bit more before everyone starts talking at once. The noise can be a bit deafening.

Everything in this PR aside from the descriptions isn't wanted. You went on a tangent and included too many things without clearing it with us first. Also, try to make each issue and pull request focused on one specific thing so the good bits don't get held back by the rest.

JeffBobbo:
I'm far too tired right now to fix my commit history into a state which you deem acceptable, I'll try to get around to it when I can to get the descriptions in.

Although could you please inform us of these kinds of conditions in the future? Most of this issue could've been avoided if you or shirepirate had said about this 24 hours ago when I originally brought it up.

PR #214, opened 2015-06-28
cannon -> blaster

freezurbern:
While the plasma isn't a cannon, it also isn't much of a revolver, nor can it use the name pistol as that is in Halo and is already used by one of the other weapons. That's why I propose we call the plasma a plasma blaster.


shirepirate:
the names aren't up for discussion and were purposely not put to a vote, if a change is to be made it will be quin's decision, but we're not welcoming suggestions for existing loadout weapons.

JeffBobbo:
@shirepirate Why are the names not up for discussion?

shirepirate:
>Why are the names not up for discussion?

some parts of development (especially creative decisions) are best left to the discretion of the creators. if they say they don't wish to put it to vote or aren't looking for suggestions it's disrespectful to continue to harp on it, that's all.

Issue #230, opened 2015-06-29
IRC Chat - No Cursor


swooboo:
When opening the IRC chat, no cursor is shown until some text is typed. Screenshot:

![20150629211512](https://cloud.githubusercontent.com/assets/900603/8415009/80470f52-1ea4-11e5-804d-b5fe664ef4bd.png)

swooboo:
Correction - this issue appears in all the text boxes.

PR #231, opened 2015-06-29
cursor now shows when there is no text (#230)

swooboo:
After diving into the code for 1 hour, I finally found where to draw the cursor, so it shows when there is no text in the textbox while it's focused. This is a solution for issue #230.


swooboo:
@qreeves can somebody review this? This may not look too good, but I've tested it and it works. Or should I earn the trust somehow?

Issue #126, opened 2015-03-26
Variable menu: checkboxes for bitwise variables.


freezurbern:
A very useful enhancement (and bringing back a part of the past) would be to have checkboxes for the variables that take a bitwise value. ~~The GUI could scan the usage descriptions if they were in a common format, finding what values should be shown.~~

qreeves:
A new parameter would probably be a better idea than trying to parse the long description.

bonifarz:
Yeah, I thought of that a while ago, but parsing descriptions will not work well here (a first attempt was quite useless). As an unofficial modification, you could add some bitfields with generic labels (e.g. as in the comment line for flags in the vars menu), but that's an ugly hack.

bonifarz:
Hey @freezurbern, here's sort of a work-around for this:
https://github.com/red-eclipse/community/blob/master/bonifarz/vars.cfg
This checks if the variable's max value takes the form 2^n -1 and displays bitfields accordingly.
Since I exclude hex (colour) values, this one is based on #239 

I guess a clean solution would still need a new flag for bitlist vars and a nicer way to show the usage instructions, but this seems to work okay for now :smiley:

bonifarz:
I think you can keep the issue open as a feature request (enhancement) in the sense that we could use a flag to identify bitlist vars and maybe better usage instructions to extract bitfield labels. But I'm glad if my workaround helps a bit.

qreeves:
@bonifarz I replaced the last field ("usage") with a general "fields" array. It uses explodelist to break down the string into pieces the same what it evaluates code (so []() etc are reserved keywords now).

Example:

`setdesc (concatword $w collide $m) "determines collision properties for a projectile from each firing action (bitwise OR)" "COLLIDE_TRACE COLLIDE_PROJ COLLIDE_OWNER IMPACT_GEOM IMPACT_PLAYER IMPACT_SHOTS BOUNCE_GEOM BOUNCE_PLAYER BOUNCE_SHOTS DRILL_GEOM DRILL_PLAYER DRILL_SHOTS STICK_GEOM STICK_PLAYER"`

Gives:
![20160129030349](https://cloud.githubusercontent.com/assets/3297929/12652047/ef2c10e2-c634-11e5-8ede-766bdc7e2960.png)

bonifarz:
Looks nice... so, lot's of changes in usage.cfg. Will have a closer look later.

qreeves:
Yeah, part of why I haven't closed this issue yet.

[...]

PR #239, opened 2015-07-01
improve vars menu: Flag filters & hex values

bonifarz:
add proper bitfields for filtering or excluding variable flags
per default, only show the bitfield for including commands
all other bitfields go in an advanced filter panel, enabled via a checkbox
check hex colours via their flag, show the default and value as such
nicer colour preview, also for default colours

### DIFF
--- a/config/menus/vars.cfg
+++ b/config/menus/vars.cfg
@@ -6,37 +6,24 @@ vartypes = 7
 varnotypes = 0
 varflags = 16
 varnoflags = 128
+varsmore = 0
+varflagnames = "persist readonly rewrite world complete texture client server hex admin unknown args preload gamepreload"
 
 newgui vars [
-    //  guilist [
-    //      guitext "varflags:  "
-    //      loop i 8 [ guibitfield $i varflags (<< 1 $i) ]
-    //      guitext "  varnoflags:  "
-    //      loop i 8 [ guibitfield $i varnoflags (<< 1 $i) ]
-    //  ] 
-    //  guibar
     guiheader "variables and commands"
     numvars = (getvarinfo -1 $vartypes $varnotypes $varflags $varnoflags $varsearchstr)
     if (>= $varnum $numvars) [ varnum = -1 ]
     guilist [
         guistrut 1
         guifield varsearchstrval 32 [varsearchstr = $varsearchstrval; varnum = -1; varindex = 0] -1 0 "" 0 "^fd<enter search terms>" 1
-        guistrut 1
+        guispring 1
         guilist [
             guibackground $guifieldbgcolour $guifieldbgblend $guifieldbordercolour $guifieldborderblend 1
             guistrut 0.15
             guilist [
-                guistrut 0.5
-                guibitfield "integer" vartypes (<< 1 0) [varnum = -1] 
-                guistrut 1
-                guibitfield "float" vartypes (<< 1 1) [varnum = -1] 
-                guistrut 1
-                guibitfield "string" vartypes (<< 1 2) [varnum = -1] 
-                guistrut 1
-                guibitfield "command" vartypes (<< 1 3) [varnum = -1] 
-                guistrut 1
-                guibitfield "alias" vartypes (<< 1 4) [varnum = -1] 
-                guistrut 0.5
+                guitext " ^fashow " 
+                guibitfield "commands  " vartypes (<< 1 3) [varnum = -1] 
+                guicheckbox "advanced filters  " varsmore
             ]
             guistrut 0.15
         ]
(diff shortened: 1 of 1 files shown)


bonifarz:
vars-preview

swooboo:
Pulled it. The default is a nice addition. But I have an issue with the **flags / exclude** column,
because it is not easily understood what the first column of checkboxes does vs the second.
This is definitely a step in the right direction though. I wish it looked more like the old vars
editor...

bonifarz:
Right, we need a better explanation, maybe with a tooltip. Suggestions are welcome.

Flags checked in the first column must be present for a var to show up in the list (i.e. like a filter, right?). Per default, this is the auto-complete flag, so user defined or temporary aliases are ignored.

The second bitfield (checkbox) includes a variable if it has the given flag. Per default, those are server vars sv_*.

Actually, the same pair of filters is available for variables types, too (vartypes/varnotypes).
I thought excluding types is not useful, but maybe the layout looks nicer with two bitfields per line.

Old vars editor: You refer to RE 1.4?

swooboo:
Yes, I am referring to the 1.4 vars editor. Also, from your explanation I still can't get
the meaning of the checkboxes, but maybe it's just me, or maybe I just didn't explain
myself. I get what the filter is for, that's self-explanatory. What I am confused about,
is the double column of checkboxes under  **flags / exclude** - maybe I'm not used
to editing vars a lot, so I just don't know what this is. As for tooltips - that'd be nice.
I actually think every thing that can be accessed with a keyboard key should have
the key in the tooltip. (Sorry for offtopic)

bonifarz:
Let me try to summarize again: There are variables *vartypes varnotypes varflags varnoflags* that allow advanced filtering when building the list of variables in the menu. These specify which variable types or flags are required or forbidden when listing the variables. That is, if I understood it correctly :wink:

qreeves:
@bonifarz Maybe you should center the text and put the exclude checkbox on the other side of it.

@swooboo We can't really maintain a menu like the old one (if anyone noticed, it got outdated very quickly during development), which is why I created the new menu which maintains itself. Perhaps in the future we could create subset pages for weapon and game vars only which presents the information in an organised list, but still uses the vartype lookups to get the information.

bonifarz:
Was just updating while you were commenting :smiley: 
As you can see, I thought of moving the advanced filter checkbox to the right, too, so it servers as sort of a title for the panel. That's how it looks now.
vars-preview
The extra checkbox for showing commands is still there, but only if the advanced panel is off.

shirepirate:
still doesn't look like you've solved the problem of the confusion with why there are two checkboxes in front of each option. that's why qreeves suggested one of them go after the text. you've removed "exclude" entirely, so it's anyone's guess as to what it means at this point.

the vars menu should be as simple and intuitive as possible while still providing advanced functionality. when you make a change you need to think like a beginner that wants to change a variable, not going into it with the assumptions and knowledge you have.

that's why the 1.4 menu was popular and several have expressed wanting a reversion to a menu like this. as it's impossible to maintain such a style of format we'll eventually need to create groups to summarize each type of variable according to it's purpose & application.

shirepirate:
also, for HEX flags (i'm going to assume these are probably all colors) it couldn't hurt to automatically add a color slider to apply a new option, similar to the profile menu

bonifarz:
> should be as simple and intuitive as possible 

Very true, that's why the panel is disabled by default, while giving a simpler option for listing commands.
> you've removed "exclude" entirely

I thought it is better to explain each checkbox via the tooltip.
> Maybe you should center the text and put the exclude checkbox on the other side of it. 

Ah, now I see. Something like this?
vars-preview
I am not sure if that looks nicer, though.

RGB-sliders for the colours? I could try that, as there is enough of space.

bonifarz:
vars-preview
I noticed I had some commented out block for issue #126. Tweaked this too, so it is somewhat useful. Easily removed if not wanted here. Also, if there's too many commits here, I can open another PR with a single commit.

qreeves:
The number of commits is unimportant, so long as it doesn't contain duplicates of commits we've made to the main repo (eg. *git pull* without *--rebase*).

qreeves:
I think a green '+' and a red/orange '-' at the top of the checkbox columns would fit the current iteration.

bonifarz:
I like that idea. Tried it with normal (a bit small) and emphasis text (hard to align). I think icons of scale 0.5 work even better. What do you think?
vars-preview
And that's how the work-around for #126 looks. Shown for any ints that take less than 20 bits. Could that be useful, or shall i discard it again, so we can look for a clean solution later?

bonifarz:
I think I found a nicer solution: Use a custom bitfield with different textures to exclude flags or types:

                        //guibitfield "" varnoflags (<< 1 $i) [ varnum = -1] 
                        guistayopen [ guiimage "textures/warning" [
                            varnoflags = (^ $varnoflags (<< 1 @i)) ; varnum = -1
                        ] 0.5 0 "" [] (? (& $varnoflags (<< 1 $i)) 0xff0000 0x999999)]

vars-preview
I just wonder, would that look better with this centered layout, or with the old, left adjusted double "checkboxes"? I guess the former is okay.

qreeves:
I dunno, I guess we'll see.

bonifarz:
Okay, thanks. If there are suggestions to further tweak the menu, it is no problem to push another PR with improvements.

Issue #241, opened 2015-07-02
Some voice-overs don't play unless followed by other characters


freezurbern:
To reproduce: Type `argh` and nothing will sound. Next, type `argh!` or `argh.` or `arghhh`
Many of the voice overs behave like this, though a notable exception is `no problem` which works with just those two words.

I think this is inconsistent, and should be changed to allow someone to say for example `thanks` and have the voice over, instead of requiring more characters.

bonifarz:
Could only reproduce with /say or /saycommand, but not with the default binds of chat keys. No clue why, though.

swooboo:
My game works like @freezurbern described. It's also weird that if I say 'suck' there is no voice
over, but 'suck!' does have the same effect as 'suck it'. Also, 'suck it' is one more exception, just
like 'no problem' that works without '!'.

PR #252, opened 2015-07-07
Fix usage of len arg in engine/server.cpp:filterstring.

zaquest:
Fixes voice overs #241. The problem was with `filter` command used in `config/voice.cfg`. The command stripped last character of a string. It did so because `engine/server.cpp:filterstring` processed one character less than the given length.


qreeves:
Actually that's an overflow. I know what to do now though, so thanks.

PR #255, opened 2015-07-08
improve bind lists in the controls menu

bonifarz:
list all relevant editing binds in the controls menu (options), so the editing menu can be overhauled later based on this.
make sure all controls are arranged nicely, e.g. ordering of F1-F12 default binds, grouping of chat, movement, weapons etc.
fix a few broken control lookups, e.g. bot compass
add a few more missing ones, e.g. help menu, server menu, waitmodeswitch ...
add fast forward for offline specators (we could put that in setup.cfg as well)
change checkbox label and default (seemed confusing)

### DIFF
--- a/config/menus/options.cfg
+++ b/config/menus/options.cfg
@@ -1165,9 +1165,8 @@ newgui options_controls [
     guilist [
         guilist [ guistrut 28 ]
         guilist [
-            guitext ( format "set basic keyboard controls from this menu. for more advanced functions, use the /bind command (%1)" (dobindsearch "saycommand /") )
-            guistrut 0.5
-            guibutton "^frrestore default controls" [showgui resetcontrols]
+            //guitext ( format "set basic keyboard controls from this menu. for more advanced functions, use the /bind command (%1)" (dobindsearch "saycommand /") )
+            //guistrut 0.5
             guistrut 0.5
             guitext "click an action then press the desired keys you wish to bind (press ESC when finished):"
             guistrut 1
@@ -1175,7 +1174,9 @@ newgui options_controls [
                 guitext "state:"
                 guistrut 1
                 loop i 4 [ guiradio (at $bindtypes $i) curbindtype $i; guistrut 1 ]
-                guicheckbox "not only state-specific binds" curbindshow
+                //guicheckbox "not only state-specific binds" curbindshow
+                guispring 1
+                guibutton "^frrestore default controls" [showgui resetcontrols]
             ]
             guistrut 1
             curbindsign = (at $bindcalls $curbindtype)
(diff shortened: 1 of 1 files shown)


config/menus/options.cfg
```diff
@@ -1143,7 +1175,7 @@ newgui options_controls [
                 guitext "state:"
                 guistrut 1
                 loop i 4 [ guiradio (at $bindtypes $i) curbindtype $i; guistrut 1 ]
-                guicheckbox "show default bind if not found" curbindshow
+                guicheckbox "not only state-specific binds" curbindshow
```
shirepirate: i glance at this and i'm not even sure what the phrasing means, how on earth is someone seeing these controls for the first time going to understand it? make it clearer if you can

bonifarz: Well, i can revert that label to how it was previously, if you dont have a better suggestion. But you agree to have it on by default, so the wait binds are not all seemingly empty? What the checkbox does is show regular binds, too, not only editbinds or specbinds. "Default binds" reminds me much of "reset to default" :smiley:

qreeves: Well, they're default in that if there's no state-specific bind it falls back to a default if found. Maybe "show fallback bind if not found"?

bonifarz:
I really don't want to be nit-picking about that checkbox, it was just one of the details I came across while working on the editing controls list.
As far as I understood, we look up actions that are useful when in spectator or waiting state, not only specbinds or waitbinds, which are very, very few. That's what can be a bit confusing with said checkbox and the way we list the binds. For example, if we look up a spectator action, we get the matching specbinds, and if the checkbox is on we also get matching (regular) binds (unless the same key is used for another specbind, of course). It does not matter if another specbind key was found for this action. But hey, that's really not important.
When I find the time, I would love to add another guitab for unused keys. The idea is to show an ordered list of (all?) keys, with some icons and/or keywords indicating if the key is still available for a bind, editbind etc. If we have such a thing, maybe we don't need said checkbox at all, and we can always show the matching keys?

qreeves:
Maybe something more akin to a list, like the variables menu?

bonifarz:
Scrollbars and filters? I don't know. But I like if the thing fits on one page, like a cheat sheet.
For the new tab idea, what about this?
getbinds
So, we readily see which keys have different binds for spec mode etc (including those funny nullbinds). I guess that helps to not override binds by mistake.
Do you think we still need a checkbox to hide any keys?

qreeves:
As another tab, I think the idea is fine. I'd say the only thing you'd have to be concerned about then is that tab making the whole UI bigger, leaving a heap of empty space in the others (my solution to that is usually to center shit).

bonifarz:
Agreed. Also depends on the number of columns in the advanced tab - four or five. I should do some more testing and also look at the thing with a different resolution.

bonifarz:
I guess I'll just open another PR based on 2df3fd7. I don't want the risk of a merge messing things up :smiley:

PR #265, opened 2015-07-11
enhance controls menu (replaces #255)

bonifarz:
revamp the list of actions and descriptions for key binds, especially for editing
add a new advanced controls tab, an overview of bind usage ordered by keys
comment out the *curbindshow* checkbox, as it seems more confusing than useful for now (do you agree?)


bonifarz:
screenshots taken from #255
 Old edit controls menu with checkbox off for comparison:
oldeditbinds
new edit bind list with checkbox and first line commented out, reset button moved:
neweditbinds
The new tab:
editbinds

Issue #281, opened 2015-07-16
editing menu overhaul


bonifarz:
Forum topic with preview images: http://redeclipse.net/forum/viewtopic.php?f=4&t=770

Summary: 
  :white_check_mark: stable branch (automatic updates)
  :white_large_square: master branch (devel version)
  :grey_question: ongoing or open for discussion

See also: https://github.com/red-eclipse/community/blob/master/bonifarz/menubackports.cfg

:white_check_mark: improve controls menu, so it serves as a key reference #255 
:white_check_mark: vars menu filters, so all world vars can be looked up #239
:white_check_mark: add new menus from forums, remove clearly unused ones
:white_check_mark: set up the glue and colour picker to start work on new edit menus
:white_check_mark: recreate v-tex menu using colour picker for vcol, glow, spec
:white_check_mark: button and checkbox macros with built-in display of edit bind, console tool-tip and console alt-action
:white_check_mark: split and recreate the editing main menu
:white_check_mark: split and recreate the editing environment menu
:white_check_mark: merge lighting and environment menus (to be expanded)
:white_check_mark: set up new editing main menu and compass
:white_check_mark: add newent icons in ents menu (work in progress)
:white_check_mark: create and merge pull request so the stuff can be tested #281
:white_check_mark: editbind E in setup.cfg, make it show and close the compass
:white_check_mark: revise editing controls in game options
:white_large_square: new tab in map info: waypoints (with file check) and routes (with counting)
:white_large_square: new tab in geometry menu: prefabs (with scrollable file list and gridsize info)
:white_large_square: replace util tab in texture menu with recreated blendmap menu 
:white_large_square: second PR for testing #292 
:white_large_square: tweak ents menu, fix sub-selection button, simpler entloop code
:white_large_square: scrollable list of sounds, enhanced mapmodel list, light presets with colour picker
:white_large_square: remove unused menus (newcolour, quick edit, teleport...)
:white_large_square: simple music menu (not only for editing)
:white_large_square: third PR for testing #294 
:white_large_square: new tabs in materials menu for properties and obits
:white_large_square: grass tab in world menu
:white_large_square: enhance sky tab in world menu with a large array of vars
:white_large_square: new submenus for picking skybox, cloudbox, cloudlayer, envlayer
:white_large_square: split geometry menu, new texturing tab 
:white_large_square: add glass2...4 to tab 2 of materials menu, a note about glass in tab 1
:white_large_square: fourth PR for testing #298
[...]

PR #298, opened 2015-07-22
editing menu overhaul, part 4

bonifarz:
another update for #281

### DIFF
--- a/config/menus/editing.cfg
+++ b/config/menus/editing.cfg
@@ -450,7 +450,6 @@
             guispring 1
             guibutton "pick" [pickcolour fogcolour] [] $editingtex $fogcolour
             guistrut 8
-            
         ]
 
         guitab lighting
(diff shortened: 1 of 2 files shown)


config/menus/editing.cfg
```diff
@@ -431,6 +435,65 @@
         guitext " "
         guibutton      "^foRESET ^fwenvironment to default"      [ambient 0 0 0; skylight 0 0 0; skybox ""; skycolour 255 255 255; cloudlayer ""; cloudlayercolour 255 255 255; cloudlayerblend 1.0; cloudscrollx 0; fogcolour 128 153 179; fog 4000; calclight -1]
 
+        guitab grass
+        if (= 0 $guipasses) [ grasslist = [
```
qreeves: You can put these outside the newgui calls, we create lists like these for the gui everywhere (eg. keybinds, etc).

bonifarz: Right.
I do this with most guitabs here to make them easier to maintain, e.g. when deciding to move a tab from one menu to another. The aliases are not local anyway, right? So technically, the only difference is an extra if-check in the gui loop, but I guess that does not kill performance.

For this particular example, I could just put the grasses on the looplist line. I'm just not sure if I want to parse data/textures to generate grass and cloud lists, just as I do with the skyboxes. 

There's nothing important in this PR, we can merge it after I have another update for the world menu.

bonifarz:
Looks fine with 1647282 - I think this update is completed for now.

PR #301, opened 2015-07-25
pickcolour: Switch default with current because

IceflowRE:
its better to see what has changed.


bonifarz:
To rephrase:
Show default, current, new colour, 
instead of current, default, new colour.

The suggested new ordering is fine for me.

bonifarz:
I like the button functionality, but did you actually revert the first commit when adding these? :wink:

PR #307, opened 2015-07-27
Sliders work if you drag outside.

Robalni:
Sliders only work when you have your mouse over them. This becomes a bit annoying when you try to use them and look at something else at the same time, like the server list. This pull request fixes that problem.


qreeves:
This actually can't be done this way due to the way the UI works; you can't generally carry states from one pass to the next. The menu could change or disappear at any moment and the list you store here would become invalid. This is the reason the slider requires the mouse to be active on it.

PR #312, opened 2015-07-28
add {duel,survivor}maxqueued variables

zaquest:
Adds two variables `duelmaxqueued` and `survivormaxqueued` that set maximum number of players that can be queued respectively for duel and survivor.

### DIFF
--- a/src/game/duelmut.h
+++ b/src/game/duelmut.h
@@ -49,11 +49,13 @@ struct duelservmode : servmode
     {
         if(ci->state.actortype >= A_ENEMY || ci->state.state == CS_SPECTATOR) return;
         if(gamestate == G_S_OVERTIME && !restricted.empty() && restricted.find(ci) < 0) return;
-        if(DSGS(maxqueued) && duelqueue.find(ci) < 0 && playing.find(ci) < 0) {
+        if(DSGS(maxqueued) && duelqueue.find(ci) < 0 && playing.find(ci) < 0)
+        {
             int player_cnt = 0;
-            loopv(duelqueue) if (duelqueue[i]->state.actortype == A_PLAYER) ++player_cnt;
-            loopv(playing) if (playing[i]->state.actortype == A_PLAYER) ++player_cnt;
-            if(player_cnt >= DSGS(maxqueued)) {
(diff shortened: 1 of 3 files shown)


src/game/vars.h
```diff
@@ -27,6 +27,9 @@ GVAR(IDF_ADMIN, serveropen, 0, 3, 3);
 GSVAR(IDF_ADMIN, serverdesc, "");
 GSVAR(IDF_ADMIN, servermotd, "");
 
+GVAR(IDF_ADMIN, duelmaxqueued, 0, 0, MAXCLIENTS); // number of players that can be queued for duel. 0 = any number of players
+GVAR(IDF_ADMIN, survivormaxqueued, 0, 0, MAXCLIENTS); // number of players that can be queued for survivor. 0 = any number of players
+
```
shirepirate: these should be placed by the other duel & survivor vars, respectively.

also they can be regular variables like the other settings, not necessary for them to be admin vars.

src/game/duelmut.h
```diff
@@ -49,6 +49,17 @@ struct duelservmode : servmode
     {
         if(ci->state.actortype >= A_ENEMY || ci->state.state == CS_SPECTATOR) return;
         if(gamestate == G_S_OVERTIME && !restricted.empty() && restricted.find(ci) < 0) return;
+        if(DSGS(maxqueued) && duelqueue.find(ci) < 0 && playing.find(ci) < 0) {
```
qreeves: Incorrect code format, braces in Cube 2 are placed on the following line.

src/game/duelmut.h
```diff
@@ -49,6 +49,17 @@ struct duelservmode : servmode
...
+            int player_cnt = 0;
+            loopv(duelqueue) if (duelqueue[i]->state.actortype == A_PLAYER) ++player_cnt;
```
qreeves: Unless you're using the value immediately, you should use: `player_cnt++`

qreeves:
I'm curious what this is needed for, why do we need to limit the queue?

shirepirate:
it provides a way for a server to have a limited amount of players without needing to be locked or have a moderator around, while still allowing others to spectate

PR #322, opened 2015-07-31
fix host msg unknown and hex colours

IceflowRE:
Fix style of a hexcolour, and if getlcienthost returns "unknown" it will show "localhost".


bonifarz:
Side note: In a commit or PR summary, I try to state *which* code or feature is addressed (e.g. tweak clients menu). Just to make it easier to see what this is about, also in the commit history. There's more room for details in the description field.

IceflowRE:
Fix style of a hexcolour, and if getlcienthost returns "unknown" it will show "localhost".

PR #325, opened 2015-08-01
Basic Statistics Output System

shacknetisp:
Here is a basic statistics output system that writes game statistics to a sqlite database of the master server.
This adds server auth handles too, so servers can only submit data if the have the 's' auth flag.

The major additions include:

* Server Auth Handles
 * These are displayed in the server browser.
* Addition of sqlite3 into the source, no new dependancies added.
* Saving of game/server/player/weapon statistics to a master server database.
 * This includes things like weapon accuracy.

This pull request is so I can be told where I need to rewrite logic or fix style, and eventually for actual merging.

I have tested this in a game, and all appeared to be working, here is a dump of the output of the database with one game (except game_weapons is missing the `playerhandle` column, I added that just after): http://ghostclanre.tk/ftp/beha/sqliteout.html 

Web Interface/API Project: https://github.com/shacknetisp/statsdb-interface


src/game/auth.h
```diff
@@ -74,6 +74,15 @@ void localopadd(const char *name, const char *flags)
...
 VAR(IDF_PERSIST, quickauthchecks, 0, 0, 1);
+
+VAR(IDF_PERSIST, authconnect, 0, 1, 1);
+SVAR(IDF_PERSIST, accountname, "");
+SVAR(IDF_PERSIST, accountpass, "");
+ICOMMAND(0, authkey, "ss", (char *name, char *key), {
+    setsvar("accountname", name);
+    setsvar("accountpass", key);
+});
+
```
qreeves: These will actually clobber the VAR/COMMAND's in `src/game/client.cpp`. You should use a different name for the server side commands.

shacknetisp:
Here is an in-progress web interface/JSON API written in Python to provide external access to the database, currently it can return the data from all the SQLite tables in JSON format.
https://github.com/shacknetisp/statsdb-interface

PR #331, opened 2015-08-02
Initialize Authhandle

shacknetisp:
masterclient.authhandle was uninitialized, probably the cause of the weirdness of the master server's output.


shacknetisp:
Fixed in 37f99708dd9f

PR #336, opened 2015-08-03
Update optionsmenu

IceflowRE:
Add
- guitooltip in advanced control menu looks now the current bind

Remove
- waterenv (not longer in use)
- lightmodels  (not longer in use)
- glowmodels  (not longer in use)

### DIFF
--- a/config/menus/options.cfg
+++ b/config/menus/options.cfg
@@ -546,6 +546,7 @@ newgui options_ui [
                 guicheckbox2 "show crosshair" showcrosshair 1
                 guicheckbox2 "weapon crosshairs" crosshairweapons 1
                 guicheckbox2 "critical flash" crosshairflash
+                guicheckbox2 "show crosshair distance" crosshairdistance
         guistrut 1
         guitext "weapon ammo/clip surrounding crosshair:"
         guilist [
@@ -640,6 +641,8 @@ newgui options_ui [
                 checkboxdesc "apply weapon color to crosshair" 1
             ] "critical flash" [
                 guitext "crosshair flashes red at critical health"
+            ] "show crosshair distance" [
+                guitext "shows distance to geometry where you are aim at"
             ] "no visible clip" [
                 guitext "no clip surrounding the crosshair is displayed"
             ] "solid bar of clip" [
@@ -656,7 +659,7 @@ newgui options_ui [
                 checkboxdesc "textual percentage" "graphical bar"
             ] "velocity meter" [
                 guitext "toggles visibility of velocity meter"
-                checkboxdesc "visible only in time race" "always visible"
+                checkboxdesc "visible only in race" "always visible"
             ] "spectator input" [
                 guitext "toggles visibility of keyboard input when watching other players"
             ] "own input" [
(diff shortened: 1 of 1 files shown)


bonifarz:
the tooltips are a good idea - are you sure you do not want to specify a width (arg2) for these?

bonifarz:
Sorry for my concerns with the tooltips, variable width is fine here. But....

Is *waterenv* no longer in use? There's still *waterenvmap* in src/engine/water.cpp - is that what the checkbox @IceflowRE removed should be?
I don't know about light and glow options for models - is it correct to remove these?

IceflowRE:
The "vars" i removed arent used by RE. (You can even save whatever you want in them) and i have take a look at waterenv, it looks like waterenv was never existing (doenst exist in 1.4.5 too) or only short in some dev versions. Waterenvmap is there yes, but i think we need no setting for this var because if its on or off, i see no difference and the fps is the same.

bonifarz:
You or me cannot really tell, let us wait what someone with better engine knowledge will say.

Another thought: As you were addressing the advanced controls tab: I realize it can imply a significant drop in frame rate, as there are **many** buttons in this one. Maybe it is better to move it to a separate menu (newgui), so you don'tget this drop in fps when using the other tabs of the controls menu?

config/menus/options.cfg
```diff
@@ -87,8 +87,6 @@ newgui options_graphics [
                 guicheckbox "^fyrefraction" waterrefract
                 guistrut 1
                 guicheckbox "^foreflection" waterreflect
-                guistrut 1
-                guicheckbox "^fgenvmap" waterenv
```
qreeves: This variable is actually now called `waterenvmap`.

config/menus/options.cfg
```diff
@@ -196,12 +194,8 @@ newgui options_graphics [
                 guiradio "^fyhigh quality" maxtexsize 0
             ]
             guilist [   // models
-                guicheckbox "^fglighting" lightmodels
-                guistrut 1
                 guicheckbox "^fgreflection" envmapmodels
                 guistrut 1
-                guicheckbox "^fgglow" glowmodels
-                guistrut 1
                 guicheckbox "^fybumpmap" bumpmodels
```
qreeves: You could *maybe* added a slider here to control `fullbrightmodels` instead.

shirepirate: >You could maybe add a slider here to control fullbrightmodels instead.
i added that to gameplay before, as there don't seem to be any other sliders in this GUI

shirepirate:
>Maybe it is better to move it to a separate menu (newgui), so you don'tget this drop in fps when using the other tabs of the controls menu?

use guivisibletab for that

IceflowRE:
The fullbrightmodels are controled in the gameplay menu already and here an guislider.. looks not very well i think.

bonifarz:
poor fps on advanced controls tooltips 
> use guivisibletab for that

Mhm, no, I think it is better to clean up the code. My advanced controls menu was a moderate waste of resources, and the tooltips just make that even worse. I'll see if storing the binds in local vars when opening the menu helps. Should be much more efficient.

shirepirate:
>I think it is better to clean up the code

definitely, but no reason to place it in a separate GUI as previously mentioned when there is guivisibletab

bonifarz:
For a work-around, guivisibletab is okay, sure.
First attempt to store binds did not help. But we could try a slider layout to reduce the number of visible buttons. *shrug*

bonifarz:
> - time race = race (update to an newer commit change)

I think this is where you may consider starting a new pull request? At least, that's what I do in such a case, although I'm not sure if merging this stuff would cause any trouble. It's just a single file to push anyway.

PR #349, opened 2015-08-08
a help menu for weapons

bonifarz:
I would like to put this draft here as PR, so it can be reviewed conveniently.
Probably this will need some more tweaks before it can be merged. 
Feedback is welcome, either here or on http://redeclipse.net/forum/viewtopic.php?f=4&t=787

weapons


bonifarz:
Too early for another PR, but here's a code update for those who are curious to test this stuff:
http://redeclipse.net/forum/viewtopic.php?f=4&t=787&p=6894#p6894

PR #354, opened 2015-08-13
improve editing menu: world

bonifarz:
move fog and background colour to its own tab, add detailed fogdome vars 
add some nice previews in the grass tab
fix line spacing in sky tab when enabling/removing layers or boxes
add a reset button for skybox/cloudlayer options, keeps the chosen sky textures
detail: vcolour --> colour in texture menu

### DIFF
--- a/config/menus/editing.cfg
+++ b/config/menus/editing.cfg
@@ -417,91 +417,6 @@
         ]
         guistrut 0.5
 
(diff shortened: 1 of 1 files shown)


bonifarz:
with the next update i should make sure that advanced lightmap options are described as such and easy to reset

PR #367, opened 2015-08-25
add hexcrc, very similar to hexcolour

bonifarz:
same as hexcolour, but for 4 byte instead of 3

Purpose: Display `(hexcrc $mapcrc)` in some menu or script


qreeves:
CRC will be replaced by hashes in the near future, so I wouldn't bother with this. Semaphore has been updated causing the builds to fail again, I'll deal with it when I get a chance.

bonifarz:
Okay, thanks for the info.
Was just a trivial addition anyway.

PR #370, opened 2015-08-28
add prefabpreview to editing menus

bonifarz:
show a large prefabpreview in the geometry menu, tab 2
use the same new layout for the mapmodel menu:
    Preview to the right of the list, using the mouse rollover item
init checkboxes for experimental shader editing feature (just fixes a console error)
init sliders on colour picker to the current value of the variable


bonifarz:
prefabs

PR #381, opened 2015-09-13
add veto warning to maps menu

bonifarz:
replace the (isonline) check with a check for other players using (getclientcount)
this fixes the menu when servertype > 0 or when playing on a server without other players
show a warning when the privs and mastermode result in a veto, so games are not forced by mistake
show a mastermode button below, but only if other players are present and you have the privs

### DIFF
--- a/config/menus/maps.cfg
+++ b/config/menus/maps.cfg
@@ -445,14 +445,39 @@ mapsmenu = [
                 guiradio "public" servertype 2
             ]
             guispring 1
-            if (>= $lastmode 0) [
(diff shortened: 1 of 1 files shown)


freezurbern:
I think the last part,
>only if other players are present and you have the privs

is a great compromise between offering a mastermode button and not doing so.

bonifarz:
It's the same idea with moderation stuff on the clients menu: Only show buttons you can actually use. Note that the veto button here will only toggle open/locked, and go back to open first if it is another mastermode.

PR #388, opened 2015-09-18
addpriv and resetpriv commands

shacknetisp:
This adds the following commands:
`addpriv <cn> <priv>`: temporarily give \<cn\> local \<priv\>, \<priv\> can be from supporter (2) to administrator (5)
`resetpriv <cn>`: reset all addprivs run on \<cn\>

This follows the pecking order, so that a moderator can add or reset supporter but not moderator or above.
Administrators and above can add or reset the administrator privilege, ignoring the normal order.

The idea is from http://redeclipse.net/forum/viewtopic.php?f=4&t=756, if commands such as `admin` or `mod` would be better than `addpriv` then I'll change it.

*This breaks protocol.*


freezurbern:
:thumbsup:

Thank you for this, and for documentation.

src/game/auth.h
```diff
@@ -157,7 +158,17 @@ namespace auth
                 }
                 copystring(ci->handle, ci->authname);
             }
-            else formatstring(msg, "\fy%s elevated to \fs\fc%s\fS", colourname(ci), privname(ci->privilege));
+            else
```
qreeves: May as well be `else if(setter) .. else`, no need for the new line and added indentation.

PR #392, opened 2015-09-21
Add skippickup to disable pickup for skipped item weapons.

### LINKED ISSUE #285: skip items entirely, skiprocket
in this enhancement, skipgrenade & skipmine would both disable the ability to even pickup a grenade or mine. not only would the on-screen dialogue not appear, but pressing action would not add them to inventory.

also, a skiprocket following the same style would be nice

shacknetisp:
Closes #285 
Also disables `skip<weap>` on explosives when in kaboom.


qreeves:
You'll need to check if the `gameent` is `player1`; the way this patch is written, it will stop bots being able to pick up items the player has marked as skipped.

PR #401, opened 2015-10-01
Show client numbers in whispers.

shacknetisp:
This will make it easier to reply to whispers by showing the client numbers in the message.

### DIFF
--- a/src/game/client.cpp
+++ b/src/game/client.cpp
@@ -1077,7 +1077,7 @@ namespace client
         if(flags&SAY_WHISPER)
         {
             if(!t) return;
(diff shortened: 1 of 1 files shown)


bonifarz:
"18 days ago" :arrow_right: bump
Seems very reasonable, and even more useful than a direct reply command (which, in some other games, is prone to another whisper catching your reply).

qreeves:
I have recently moved house and it could be a while before I have an internet connection. Bumping stuff does not get it done faster; I get it done when I get it done.

bonifarz:
Of course. Just wanted to make sure this does not get forgotten - which is nonsense, as there's just one open PR anyway.

PR #412, opened 2015-12-22
Prefer clients with lower ping to host bots.

shacknetisp:
Bot hosts are currently sorted by the number of bots they own, this pull request uses ping in the sort, preferring less lagging hosts.

It uses this new formula to compare: `(hosted_bots * 50) + ping`


qreeves:
Worth an experiment, especially with the vars to tweak on the fly.

PR #415, opened 2015-12-27
player model options

bonifarz:
I'll put this here in case that's easier to review.
Discussion: http://redeclipse.net/forum/viewtopic.php?f=4&t=844

playermodels


shirepirate:
merge it with "gameplay"
would also like @qreeves input on whether a preview is necessary for these options. in my opinion it is overly confusing and not needed. if we scrap the preview you don't even need to put these options in a new tab under gameplay, just add them to the bottom.

bonifarz:
As noted on the forums, I do think the preview is essential in this context, however, I am not sure if this radio approach works. It cannot be initialized according to custom settings of player*tone vars. I'm experimenting with different approaches, which may lead to another PR. Closing this for now.

PR #420, opened 2015-12-28
add player tone vars to usage.cfg

bonifarz:
Those vars are confusing when undocumented - I hope these descriptions are correct.
Also: Inventorytone is a hex colour.


bonifarz:
If there's a mistake here, we can still correct it later

PR #424, opened 2016-01-02
Fix overflow caused by demomaxsize. Update demomaxsize usage doc.

zaquest:
Currently usage doc says demomaxsize is in kilobytes, but it's treated as if it's in megabytes and max value for the variable is too big and causes crash if set to big values. Stable is also affected.


qreeves:
Would be better to just change the var to represent bytes instead of megabytes.

PR #433, opened 2016-01-18
Run cubescript on every game start, and allow the server to run world cubescript on the client.

shacknetisp:
Some servers used map cfg files to run cubescript, such as for menus. This would allow servers to define the cubescript in `onclientmapload` instead of inside the config, making sharing it between maps much easier.

This also adds `ongamestart`, which is executed by the server on every game start, and can be used for changing server variables depending on the map that was loaded.

Two server commands were added for `ongamestart`:

* `getmapname`: Returns the map name
* `srvout <relay level> <message>`: Sends a server message.


IceflowRE:
nice next step for multiplayer adventure maps :D now are only saving to files and forcing players to spawn at specific playerstarts missing ;)

qreeves:
I can't help but feel this is half an idea at the moment, and some of it seems like a pretty weird way of doing something simple. Can you provide a more detailed description/example scenarios?

shacknetisp:
One possible application for `ongamestart` is providing various variable sets depending on the map loaded, you could have "Race CTF" for a certain set of maps, "Sumo" for another set, and normal variables for others. `onclientmapload` could then be used to display information about the game mode.

Going further, with the addition of allowing custom server cubescript to be executed by the client, it would be possible to create servers with more advanced map logic, such as hunting for hidden map triggers or solving puzzles.

IceflowRE:
Whats going on with this Pull Request?

qreeves:
OK so while I do see the point of this PR, I still think it is an idea that should be subject to more scrutiny. If we're going to create some server/client side logic we need to prepare a better "API" for doing so. Is cubescript the best option? What are the security implications of allowing servers to do this even in a "world" context? I don't think two triggers constitutes a full idea for the moment, and I believe that we should involve the community more before proceeding on this one, to find out what exactly it is people want out of this.

PR #439, opened 2016-01-23
Add variable to reset checkpoints upon material deaths.

shacknetisp:
This adds the variable `checkpointspawn` which, when disabled, causes all material deaths to behave like `suicide` and reset checkpoints.

This will solve the small inconvenience of pressing K after respawning at a checkpoint.


IceflowRE:
But it does not reset impulse bar in race endurance when suicide?

shacknetisp:
It behaves as if you pressed K, impulse is not reset in endurance.

bonifarz:
Ah, this can be nice to have... I remember @zaquest once asked if there is a variable for exactly this.

PR #449, opened 2016-02-14
Track actual weapon that fired a projectile.

shacknetisp:
This allows stats to be collected properly for weapons (like mines) that use other weapon's flak for their fragweap.


qreeves:
I don't recommend using boolean values when using network commands like `addmsg` and `sendf`. I'd probably make `fromsecondary` a `fromflags`, so you can query all the parts of the previous firing mode and make it a useful integer.

PR #459, opened 2016-03-05
enhance levelcolour (darker or brighter)

bonifarz:
- new defaults and limits for the level variables: 0.0 = original colour,  -1.0 = blended to black, +1.0 = blended to white
- previously, it just scaled the rgb vector, with a multiplier between 0 (black) and 10.0


bonifarz:
okay, that was nonsense - sorry for the noise.

PR #498, opened 2016-05-22
Send total scores from the master server's statistics database to servers.

### LINKED ISSUE #494: Auth stats integration
We could improve overall ranking of an individual integrating the new `server::savedscore` components `totalpoints`, `totalfrags`, and `totaldeaths` (and possibly even `timeactive`, `timealive`, and `timeplayed`) into the statistics database for each registered handle, then when a user authenticates on a server for the first time, the master server could send this back along with the success message to have it restored locally on the server for team balancing purposes (and then transmitted back at the end of each match with the regular payloads [if stats are enabled that is]).

shacknetisp:
When a player authenticates, the master server will send totals from the player's last `masterbalancegames` number of games. The server will use these totals in balancing.

These things may need work or comment before merging:

* Players who have played the full `masterbalancegames` will have a much larger total than an unauthed player or new player. One solution to this is to weight the totals according to how many games were played.
* The `masterbalancegames` value itself was added so that only the player's relatively recent skill would be used in balancing, very old games would not affect it. There may be a better value for this, currently it is `200`.
* Each handle's totals information is retrieved the first time a player auths on a server, and that server will keep the same information until restart. This means that once a player has authed, only the previous totals and totals from that server will be counted there, games on other servers will not be counted until the server restarts. This needs changed to allow updating the totals information while not doubly counting games in the statistics that were submitted by the server.

Closes #494

### DIFF
--- a/sql/stats/create.sql
+++ b/sql/stats/create.sql
@@ -9,7 +9,8 @@ CREATE TABLE games (
     mode INTEGER,
     mutators INTEGER,
     timeplayed INTEGER,
-    uniqueplayers INTEGER
+    uniqueplayers INTEGER,
+    usetotals INTEGER
 );
 
 CREATE TABLE game_servers (
--- a/sql/stats/upgrade_4.sql
+++ b/sql/stats/upgrade_4.sql
@@ -1,0 +1,2 @@
+/* DB Version 4 to Version 5 */
+ALTER TABLE games ADD COLUMN usetotals INTEGER DEFAULT 0;
--- a/src/engine/master.cpp
+++ b/src/engine/master.cpp
@@ -10,7 +10,7 @@
 #include <enet/time.h>
 #include <sqlite3.h>
 
-#define STATSDB_VERSION 4
+#define STATSDB_VERSION 5
 #define STATSDB_RETRYTIME (5*1000)
 #define MASTER_LIMIT 4096
 #define CLIENT_TIME (60*1000)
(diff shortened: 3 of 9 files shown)


qreeves:
I've taken some time to think about this one, mostly because it isn't exactly what I was after. While I do see the value in limiting the range for ranking purposes, I wanted to start collecting more "ongoing progress" stuff that could be tied to awards, etc. As far as a metric goes, maybe it is better to "overcome" your previous low ranking by playing better consistently rather than relying on a "reset switch"? I don't think we can know that *yet*.

I'm not sure I like the split between local and global either, especially with the way it spams data values that don't change (the global ones) with every state update. I understand the goal here is to enable these rankings to carry somewhat on servers that don't send stats themselves, but I think the only outward facing stuff should be the "total" vars the way I've already defined them. What you'd probably want to do is keep a running total of points earned on just those servers which aren't stats enabled for use in savedscore (which itself could probably use some sort of backup mechanism between server restarts, but that is a different topic).

The way I see it, the "total" vars represent "global plus local-(if not stats enabled)" and "local" doesn't need synchronising to anything but the stats server (if it is allowed to do so, at which point you could reset the running total on that server). These values would be updated on every connection to (a new game on) the server, not just the initial ones (because it is important to keep updating these totals, especially when a user first starts playing); so you get your global score, then added to that is the local score that's being tracked in savedscore (if the server *isn't* stats enabled), and that is your total one.

For now, I think I'd rather it done this way rather than trying to complicate it early on with no solid data.

shacknetisp:
Superseded by #509.

PR #509, opened 2016-06-14
Send totals statistics to servers for balancing.

### LINKED ISSUE #494: Auth stats integration

shacknetisp:
Here is a reworked, simpler version of #498.
Servers request totals from the master server upon authentication and map change.
Local scores are kept for servers that do not send statistics.

Closes #494

### DIFF
--- a/src/game/server.cpp
+++ b/src/game/server.cpp
@@ -508,6 +508,9 @@ namespace server
             ci->localtotalpoints = localtotalpoints;
             ci->localtotalfrags = localtotalfrags;
             ci->localtotaldeaths = localtotaldeaths;
+            ci->totalpoints = localtotalpoints;
+            ci->totalfrags = localtotalfrags;
+            ci->totaldeaths = localtotaldeaths;
             ci->spree = spree;
             ci->rewards[0] = rewards;
             ci->timeplayed = timeplayed;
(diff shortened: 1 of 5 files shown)


src/game/server.cpp
```diff
@@ -3288,6 +3298,13 @@ namespace server
                 }
             }
             if(!worthy) return;
+
+            loopv(clients) {
```
qreeves: ^ Formatting error.

src/game/server.cpp
```diff
@@ -3311,7 +3328,8 @@ namespace server
                     }
                 }
             }
-            requestmasterf("stats game %s %d %d %d %d\n", escapestring(smapname), gamemode, mutators, gamemillis/1000, unique);
+            requestmasterf("stats game %s %d %d %d %d %d\n", escapestring(smapname), gamemode, mutators, gamemillis/1000, unique, m_usetotals(gamemode, mutators));
```
qreeves: We usually implicitly define an integer rather than relying on bool-to-int conversion in a format string.
`m_usetotals(gamemode, mutators) ? 1 : 0`

Issue #533, opened 2016-07-05
textkeyimages 0 renders incorrectly


FaTony:
Screenshot here: http://redeclipse.net/forum/viewtopic.php?f=7&t=1000

zaquest:
I have similar issue, but for me it's because of missing textures for some keys in `data/textures/keys`. For example textures for keys `ralt`, `lalt`, `leftbracket`, `rightbracket` are missing. There's texture for `return` so it might be something different.

PR #537, opened 2016-07-06
Fix background width for #533

zaquest:
http://redeclipse.net/forum/viewtopic.php?f=7&t=1000#p9120 I'm not sure if this is an ok solution, maybe there should be a separate function to format key for output instead of spreading it between several functions.


qreeves:
> maybe there should be a separate function to format key for output instead of spreading it between several functions.
Probably.

PR #540, opened 2016-07-06
Fix zoom jerking. Closes #518. Remove unused inzoomswitch().

### LINKED ISSUE #518: Rifle jerks if unzoomed while zooming.
The rifle's zoom screen jerks to full zoom if unzoomed while zooming, making a rather confusing flash of the zoom texture.

zaquest:
`bool prevzoom` was unused too, so I thought it's a good name for the new usecase. There's a lot of repetition there probably could be something like `progress` function/macro.


qreeves:
The easier way is to just subtract the difference from `lastzoom` when it is set.

Issue #546, opened 2016-07-07
Make textkeyimages 0 cyan again


FaTony:
The button names were cyan in previous versions. Now they are white which makes them harder to read.

PR #549, opened 2016-07-08
Use cyan colour for textkeyimages 0 (#546)

IceflowRE:
Issue: #546


qreeves:
I think only the name of the key is supposed to be cyan, not the whole string.

IceflowRE:
PS.

We need more code documentation....

PR #554, opened 2016-07-12
two small mem leaks fixes :

devnexen:
lookup might have been allocated in the invalid block.
addline already allocate memory.

PS : sorry for the previous PR, was not really intended.

### DIFF
--- a/src/engine/command.cpp
+++ b/src/engine/command.cpp
@@ -1245,7 +1245,7 @@ done:
     }
     return;
 invalid:
-    if (lookup) delete[] lookup;
+    delete[] lookup;
     switch(ltype)
     {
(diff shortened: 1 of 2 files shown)


graphitemaster:
you don't have to check for null pointer before deleting

qreeves:
> PS : sorry for the previous PR, was not really intended.

Live and learn, congrats on your first successful PR. Another tip: try to avoid "merge commits" by using "--rebase".

devnexen:
good points ;) yes I just appreciate video games which work without heavy patching on BSD ;)

Croydon:
On the other hand it would have been possible to remove the "create license" commit from the other PR and don't be in the need for a new PR :)

PR #566, opened 2016-08-07
Remove limitation of guilistslider

IceflowRE:
At the moment only list are possible with ascending numbers. This change remove this limitation.

With help of @TheAssassin.

### DIFF
--- a/src/engine/menus.cpp
+++ b/src/engine/menus.cpp
@@ -535,10 +535,9 @@ void guilistslider(char *var, char *list, char *onchange, int *reverse, int *scr
         list += strspn(list, "\n\t ");
     }
     if(vals.empty()) return;
-    int val = getval(var), oldoffset = vals.length()-1, offset = oldoffset;
-    loopv(vals) if(val <= vals[i]) { oldoffset = offset = i; break; }
+    int val = getval(var), index = vals.find(val), offset = index;
     cgui->slider(offset, 0, vals.length()-1, *colour >= 0 ? *colour : 0xFFFFFF, intstr(val), *reverse ? true : false, *scroll ? true : false, *style, *scolour);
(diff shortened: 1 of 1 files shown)


IceflowRE:
I think the offset var can be removed too. I will take a look at the evening.

IceflowRE:
Similiar it would be for `guinamelist`:

```
void guinameslider(char *var, char *names, char *list, char *onchange, int *reverse, int *scroll, int *colour, int *style, int *scolour)
{
    if(!cgui) return;
    vector<int> vals;
    list += strspn(list, "\n\t ");
    while(*list)
    {
        vals.add(parseint(list));
        list += strcspn(list, "\n\t \0");
        list += strspn(list, "\n\t ");
    }
    if(vals.empty()) return;
    int val = getval(var);
    int index = vals.find(val);
    int offset = index;
    char *label = indexlist(names, offset);
    cgui->slider(offset, 0, vals.length()-1, *colour >= 0 ? *colour : 0xFFFFFF, label, *reverse ? true : false, *scroll ? true : false, *style, *scolour);
    if(offset != index) updateval(var, vals[offset], onchange);
    delete[] label;
}
```

Not tested.

src/engine/menus.cpp
```diff
@@ -535,10 +535,11 @@ void guilistslider(char *var, char *list, char *onchange, int *reverse, int *scr
         list += strspn(list, "\n\t ");
     }
     if(vals.empty()) return;
-    int val = getval(var), oldoffset = vals.length()-1, offset = oldoffset;
-    loopv(vals) if(val <= vals[i]) { oldoffset = offset = i; break; }
+    int val = getval(var);
+    int index = vals.find(val);
+    int offset = index;
```
qreeves: Squash these onto one line.

PR #580, opened 2016-09-13
Force filtered-out random weapons to be selected if needed to fill maxcarry.

shacknetisp:
Currently the random weapon filter can cause less than `maxcarry` slots to be filled. This does not seem proper behaviour to me, this PR changes it so that `maxcarry` slots are always filled, and the random weapon filter takes a lower priority than filling all the slots.

### DIFF
--- a/src/game/game.h
+++ b/src/game/game.h
@@ -838,10 +838,8 @@ struct clientstate
         }
         if(AA(actortype, maxcarry) && m_loadout(gamemode, mutators))
         {
-            int n = 0;
-            int unfiltered = 0;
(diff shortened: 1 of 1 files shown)


src/game/game.h
```diff
@@ -839,6 +839,11 @@ struct clientstate
         if(AA(actortype, maxcarry) && m_loadout(gamemode, mutators))
         {
             int n = 0;
+            int unfiltered = 0;
```
qreeves: I prefer it if same type values are grouped on the same line when possible.

PR #583, opened 2016-09-19
Add combined kdratio for better team balancing.

shacknetisp:
Currently the default balance style is by kills/deaths from global statistics. This does not take into account current performance, which means that players who are not up to their usual skill, or skilled players who do not have many games in statistics cannot be counted correctly.

This pull request adds a "combined kdratio" option that adds the k/d ratio from the global statistics with the ratio from the current game, allowing for balancing to fit the current game more closely.

### DIFF
--- a/src/game/game.h
+++ b/src/game/game.h
@@ -804,6 +804,11 @@ struct clientstate
         return frags >= deaths ? (frags/float(max(deaths, 1))) : -(deaths/float(max(frags, 1)));
     }
 
+    float combinedkdratio()
+    {
+        return kdratio() + kdratio(false);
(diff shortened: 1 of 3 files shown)


qreeves:
Probably worth noting that "usage.cfg" is way out of date with regard to "teambalance" values, eh @bonifarz ?

bonifarz:
Hah, I see https://github.com/red-eclipse/base/blob/master/src/game/vars.h#L288-L296, and we could just use these comments as descriptions, but looking at the 6 values for teambalance, it is not that trivial to understand what this means :smiley:

PR #587, opened 2016-09-26
ctf fixes

pocak100:
Please see http://redeclipse.net/forum/viewtopic.php?f=4&t=1010 for context.

### Protect ctf
Commit 9f3d1f3c forgot to account for protect ctf, which has no goal flag.

### Multiple capture glitch
When a player is near a loose flag, the client keeps sending a N_TAKEAFFIN messages every frame. When the server sees the first one, it assigns the flag to the player. If said player is also at her home base, the server gives her a point and returns the flag to its base. Then, the next queued up N_TAKEAFFIN arrives, and the server thinks the player picked up the flag again, and immediately scores again...
This patch will make the server reject N_TAKEAFFIN from the same player for one second after a capture (for the same flag only).

### Flag reset change
There are two issues here.

One is that with at least two players, a flag dropped into deathmat resets back to base instead of where it was dropped from, as intended. That's because loose flags are reset when half of the connected clients (rounded down) send an N_RESETAFFIN message. The first reset puts the flag to where it was dropped from (`floorpos`). But then, immediately, the other half of the votes roll in, and the flag is sent to its base. Raising the required number of votes did not solve this issue, because for some reason the clients occasionally double vote for each flag, usually when multiple are dropped at the same time.

An unrelated problem (e.g. happens even in offline single player) is that if a player walk off a cliff and drops a flag in the abyss, it returns to the edge of the cliff. And if that edge happens to be sloped down towards the chasm, then the flag bounces right back down into the deep. Surely, some players would learn the spots where this is reliably doable (f.e. on *Darkness* next to the pusher), and abuse it.

To solve these, this patch makes flags freeze at the gound position where they are reset to. Since `floorpos` is always at ground level and reachable by players, there's no need to apply bouncy physics. Also, since they cannot move, they cannot get in deathmat, so further N_RESETAFFIN can be ignored.

The freezing is achieved by setting the `stuck` member to 1. Also, `spherecheck` had to be disabled for stuck flags, because it was moving them off of cliffs. I hope this doesn't break stuff.

### DIFF
--- a/src/game/capture.cpp
+++ b/src/game/capture.cpp
@@ -495,14 +495,19 @@ namespace capture
         capturestate::flag &f = st.flags[i];
         if(value > 0)
         {
-            affinityeffect(i, T_NEUTRAL, f.droploc, f.above, 3, "RESET");
+            affinityeffect(i, T_NEUTRAL, f.droploc, value == 2 ? pos : f.above, 3, "RESET");
             game::spawneffect(PART_SPARK, vec(f.pos()).add(vec(0, 0, enttype[AFFINITY].radius*0.45f)), enttype[AFFINITY].radius*0.25f, TEAM(f.team, colour), 1.5f);
             game::spawneffect(PART_SPARK, vec(f.pos()).add(vec(0, 0, enttype[AFFINITY].radius*0.45f)), enttype[AFFINITY].radius*0.25f, 0xFFFFFF, 1.5f);
             game::spawneffect(PART_SPARK, value == 2 ? pos : vec(f.spawnloc).add(vec(0, 0, enttype[AFFINITY].radius*0.45f)), enttype[AFFINITY].radius*0.25f, TEAM(f.team, colour), 1.5f);
             game::spawneffect(PART_SPARK, value == 2 ? pos : vec(f.spawnloc).add(vec(0, 0, enttype[AFFINITY].radius*0.45f)), enttype[AFFINITY].radius*0.25f, 0xFFFFFF, 1.5f);
             game::announcef(S_V_FLAGRESET, CON_SELF, NULL, true, "\fathe %s flag has been reset", game::colourteam(f.team, "flagtex"));
         }
-        if(value == 2) st.dropaffinity(i, pos, vec(0, 0, 1), lastmillis);
+        if(value == 2)
+        {
+            st.dropaffinity(i, pos, vec(0, 0, 1), lastmillis);
+            f.proj->stuck = 1;
+            f.proj->stick = NULL;
+        }
         else st.returnaffinity(i, lastmillis);
     }
 
--- a/src/game/capture.cpp
+++ b/src/game/capture.cpp
@@ -517,7 +517,7 @@ namespace capture
         float radius = enttype[AFFINITY].radius;
         vec abovegoal, capturepos, returnpos;
         capturestate::flag &f = st.flags[relay];
-        if (st.flags.inrange(goal))
+        if(st.flags.inrange(goal))
         {
             capturestate::flag &g = st.flags[goal];
             abovegoal = g.above;
--- a/src/game/capture.cpp
+++ b/src/game/capture.cpp
@@ -508,14 +508,25 @@ namespace capture
 
     void scoreaffinity(gameent *d, int relay, int goal, int score)
     {
-        if(!st.flags.inrange(relay) || !st.flags.inrange(goal)) return;
+        if(!st.flags.inrange(relay)) return;
+        float radius = enttype[AFFINITY].radius;
+        vec abovegoal, capturepos, returnpos;
         capturestate::flag &f = st.flags[relay];
-        capturestate::flag &g = st.flags[goal];
-        affinityeffect(goal, d->team, g.above, f.above, 3, "CAPTURED");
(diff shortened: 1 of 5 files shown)


src/game/capture.h
```diff
@@ -192,6 +194,15 @@ struct capturestate
         destroy(i);
 #endif
     }
+    
+    bool cantake(int i, int cn, int t)
+    {
+#ifdef GAMESERVER
+        flag &f = flags[i];
+        if (f.lastowner == cn && f.returntime + 1000 >= t) return false;
```
qreeves: `1000` should probably be a game variable.

src/game/capture.h
```diff
@@ -192,6 +194,15 @@ struct capturestate
         destroy(i);
 #endif
     }
+    
+    bool cantake(int i, int cn, int t)
+    {
+#ifdef GAMESERVER
```
qreeves: Is this necessary? You could set `returntime` on the client side as well when they receive those events and have them not request at all.

src/game/capture.cpp
```diff
@@ -495,27 +495,43 @@ namespace capture
...
+        float radius = enttype[AFFINITY].radius;
+        vec abovegoal, capturepos, returnpos;
         capturestate::flag &f = st.flags[relay];
-        capturestate::flag &g = st.flags[goal];
-        affinityeffect(goal, d->team, g.above, f.above, 3, "CAPTURED");
-        game::spawneffect(PART_SPARK, vec(g.spawnloc).add(vec(0, 0, enttype[AFFINITY].radius*0.45f)), enttype[AFFINITY].radius*0.25f, TEAM(g.team, colour), 1.5f);
-        game::spawneffect(PART_SPARK, vec(g.spawnloc).add(vec(0, 0, enttype[AFFINITY].radius*0.45f)), enttype[AFFINITY].radius*0.25f, 0xFFFFFF, 1.5f);
-        game::spawneffect(PART_SPARK, vec(f.spawnloc).add(vec(0, 0, enttype[AFFINITY].radius*0.45f)), enttype[AFFINITY].radius*0.25f, TEAM(f.team, colour), 1.5f);
-        game::spawneffect(PART_SPARK, vec(f.spawnloc).add(vec(0, 0, enttype[AFFINITY].radius*0.45f)), enttype[AFFINITY].radius*0.25f, 0xFFFFFF, 1.5f);
+        if (st.flags.inrange(goal))
```
qreeves: Inconsistent formatting, no spaces between if/else/etc and the bracket.

src/game/capturemode.h
```diff
@@ -125,12 +126,15 @@ struct captureservmode : capturestate, servmode
...
         if(f.votes.length() >= int(floorf(numclients()*0.5f)))
         {
-            if(G(captureresetfloor) && f.floorpos != vec(-1, -1, -1))
+            if(G(captureresetfloor))
             {
-                ivec pos = ivec(vec(f.floorpos).mul(DMF));
-                capturestate::dropaffinity(i, f.floorpos, vec(0, 0, 1), gamemillis);
-                f.floorpos = vec(-1, -1, -1); // in case its a fail
-                sendf(-1, 1, "ri3i3", N_RESETAFFIN, i, 2, pos.x, pos.y, pos.z);
+                if (f.floorpos != vec(-1, -1, -1))
```
qreeves: Inconsistent formatting.

src/game/capturemode.h
```diff
@@ -125,12 +126,15 @@ struct captureservmode : capturestate, servmode
...
+                {
+                    ivec pos = ivec(vec(f.floorpos).mul(DMF));
+                    capturestate::dropaffinity(i, f.floorpos, vec(0, 0, 1), gamemillis);
+                    f.floorpos = vec(-1, -1, -1);
+                    sendf(-1, 1, "ri3i3", N_RESETAFFIN, i, 2, pos.x, pos.y, pos.z);
+                }
```
qreeves: I'm wondering if not having the fallback for returning directly to base is a good idea or not. What are the consequences of this change?

src/game/projs.cpp
```diff
@@ -678,6 +678,7 @@ namespace projs
     bool spherecheck(projent &proj, bool rev = false)
     {
         if(!insideworld(proj.o)) return false;
+        if (proj.stuck) return false;
```
qreeves: Inconsistent formatting.

src/game/capture.h
pocak100: The client stops sending requests as soon as it knows it got the flag (e.g. one ping time after sending the first). The problem is that at that point there are already multiple such requests traveling through the Internet toward the server. An additional check wouldn't change anything.

Anyway, I just realized that in the last version of my patch, `cantake()` had only been used in one place only, so I folded it into `captureservmode::takeaffinity()`.

src/game/capture.h
pocak100: Ok. I'm no good at naming, so feel free to rename it.

I'd like to point out that as long as it's higher than all players' ping and lower than the time it takes to go from one base to another, it should have no effect on gameplay. That's why I thought it didn't need a variable.

src/game/capture.cpp
pocak100: Thanks, fixing.

src/game/capturemode.h
pocak100: There's still `captureresetdelay` to return them. Since flags won't move after a reset, they won't get in weird or unreachable places. In what situation could the fallback be necessary?

Issue #593, opened 2016-10-17
ctf - enemy flags in deathmat don't reset to base


JeffBobbo:
Commit 78ada55 (ctf fixes (#587)) introduced some good fixes, but it seems it introduced a bug.

If I drop the enemies team's flag into deathmat, it resets to the last position much like my own flag. Enemy flags into deathmat should reset to their base.

A demo is included, first instance is at the 7 minute mark (I kill Some Assassin with a rocket, the flag falls into deathmat), another instance occurs at 2:05 (I kill zaquest and the flag bounces off the bridge into the deathmat), last instance is around the 1 minute mark.

[enemyreset.zip](https://github.com/red-eclipse/base/files/533576/enemyreset.zip)

qreeves:
Not sure if that's really a bug more than a feature suggestion. Part of the reason is the mechanics deliberately make you hand deliver your flag back to your base (that is, without the *quick* mutator). If anything, this is just enforcing this mechanic even more.

JeffBobbo:
The only times when someone would drop the enemy flag in deathmat is to troll their team mates, because the messed up (parkour) or because they died. It seems counter-intuitive to enforce someone to wait around for the flag to reset so they can take it and go.

It also devalues the worth of weapons with stun/hitpush, flamer2 and plasma2 can't be used as effectively to mess with people's parkour to drop the flag as they're running. You're also locking defenders potentially into a scenario where they can't shoot if their flag is over deathmat, since they might be dead by the time it's reset (and another enemy can take it the rest of the way).

Also, lastly I'd say note TheAssassin's and zaquest's reaction in the demo of "wtf" when it happened.

TheAssassin:
Aside this discussion, changing such a central behavior on a single server is not a good idea. This should be reverted IMO until this issue is closed.

pocak100:
Those are good arguments, @JeffBobbo. I could add a setting for `captureresetfloor` that does what you suggest. But won't it be confusing to new players?

Let's consider *multi-ctf*, too. Shooting an enemy holding a third team's flag near deathmat would become counterproductive. Is that ok?

JeffBobbo:
Perhaps, but if I'm near death I can just jump off the map and use that to buy time as it is now.

>Let's consider multi-ctf, too. Shooting an enemy holding a third team's flag near deathmat would become counterproductive. Is that ok?

[...]

PR #596, opened 2016-10-18
Extend captureresetfloor.

### LINKED ISSUE #595: flags can reset to unstable positions (and then sit on deathmat and not reset)
Flags can reset to a position where they're not stable and will fall again.
If the flag falls again, it'll sit on the death mat and not reset, waiting until the timeout to reset (although it can be retrieved by a player occasionally and force it to reset).

Example with omega flag at the 8:30 ish mark while contesting for the rocket (me, yeatta and beha).

[flagdeath.zip](https://github.com/red-eclipse/base/files/534213/flagdeath.zip)

pocak100:
The new value, 2, makes the carrier's own flag reset to where it was
dropped from, while enemy flags reset to their base. Implements #593.

The default value for `captureresetfloor` is set to 0 for now
to be compatible with older clients. Fixes #595.

### DIFF
--- a/src/game/capture.h
+++ b/src/game/capture.h
@@ -13,7 +13,7 @@ struct capturestate
         vec droploc, inertia, spawnloc;
         int team, yaw, pitch, droptime, taketime, dropoffset;
 #ifdef GAMESERVER
-        int owner, lastowner, returntime;
+        int owner, lastowner, lastownerteam, returntime;
         vector<int> votes;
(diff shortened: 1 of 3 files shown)


qreeves:
I know it is easier to think in terms of `&&` than `||` but you shouldn't use boolean `!` like that in shared sources as it can be difficult to parse mentally when speed reading over the code.

Issue #555, opened 2016-07-12
prefab to clipboard not working in coop


Dziq84:
Quinie provided me an good explanation why this is not working. Basically copying only works with known geometry. I'ts based on selection and sends that data to other clients.  I think that selecting an prefab to be pasted to clipboard should work with other clients, hope its not so hard to make this possible since its another great way to enhance coop editing and allow players to present their hard created/nice geometry :)

PR #597, opened 2016-10-19
Coop editing

zaquest:
This is a work towards #555.
- First commit fixes a bug similar to the one in #548, but regarding undo/redo. Because N_UNDO/N_REDO were sent out of order it was for example possible to introduce a situation where everyone except undoing client has a piece of geometry, since N_UNDO is sent for an action that hasn't been delivered to other clients yet. One example to reproduce it: /editbind N [ paste; undo ], then copy a piece of geometry (small enough for coop undo), press N, your client has no new geometry, other clients do have it.
- Second commit makes it possible for a player to join coop-editing game and paste a piece of geometry from a different map.
- Third commit replaces `pasteprefab` function with more granular `copyprefab`, that copies a prefab into player's clipboard. It is easy to implement `pasteprefab` in terms of `copyprefab`, but implementing `copyprefab` in terms of `pasteprefab` required double pasting prefab, deleting geometry, copying, undoing (that's when I found out about the bug fixed in 1st commit), which requires a lot of unnecessary network communications in coop editing, as well as makes it impossible to have big prefabs inserted on `V` in coop editing.

The protocol isn't changed at all, meaning clients using current version can receive and see pasted geometry/prefabs, but can't paste them themselves.

### DIFF
--- a/src/game/client.cpp
+++ b/src/game/client.cpp
@@ -1518,7 +1518,8 @@ namespace client
             outbuf = NULL;
             inlen = outlen = 0;
             needclipboard = -1;
-        } else needclipboard = 0;
+        }
+        else needclipboard = 0;
         packetbuf p(16 + outlen, ENET_PACKET_FLAG_RELIABLE);
         putint(p, N_CLIPBOARD);
         putint(p, inlen);
--- a/src/game/client.cpp
+++ b/src/game/client.cpp
@@ -1517,14 +1517,14 @@ namespace client
         {
             outbuf = NULL;
             inlen = outlen = 0;
-        }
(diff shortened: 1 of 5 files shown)


bonifarz:
Looks good - and much nicer than the cubescript hack in my menus

src/game/client.cpp
```diff
@@ -1517,14 +1517,14 @@ namespace client
         {
             outbuf = NULL;
             inlen = outlen = 0;
-        }
+            needclipboard = -1;
+        } else needclipboard = 0;
```
qreeves: Place on new line.

PR #619, opened 2016-12-06
Assign initial teams from a pool of all applicable players.

shacknetisp:
This is the first step in implementing @JeffBobbo's ideas found in [this forum post](http://redeclipse.net/forum/viewtopic.php?f=4&t=934&p=8383).
At the beginning of the game, it gathers all waiting players into a pool, then assigns them in descending balance score order to the team with the least total balance score.
I'd like to hear the thoughts of @JeffBobbo and @qreeves on this, hopefully I can continue by enhancing mid-game balance as well.


JeffBobbo:
The first team should be random, if it isn't currently, so that one player (the best) isn't always favoured by alpha.

qreeves:
Looks OK to me, so I merged it. Guess we'll see how well it does on the official server.

PR #631, opened 2016-12-16

TheAssassin:
This commit introduces the Red Eclipse URL scheme that can be used to start
Red Eclipse and immediately connect to a Red Eclipse server without having to
type a connect command or using the server browser to find a specific server.

redeclipse://hostname:port
redeclipse://hostname

Slashes appended to the URL are ignored, hostname and port arguments are
optional.

The parser uses the same functionality as the /connect command, which means
it provides similar behavior.

Beware you need to configure your system to launch the Red Eclipse client
binary when redeclipse:// URLs are opened (e.g. in a Browser or a desktop
shortcut). You can find an example for Linux desktops in the community
repository (https://github.com/red-eclipse/community).

### DIFF
--- a/src/engine/main.cpp
+++ b/src/engine/main.cpp
@@ -1056,6 +1056,38 @@ int main(int argc, char **argv)
     localconnect(false);
     resetfps();
 
+    // redeclipse:// URI support
+    // example: redeclipse://hostname:port/password
+    // (password is optional)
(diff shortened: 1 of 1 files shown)


TheAssassin:
Closing as this has become a pretty mess and I gave up on this approach after some discussions. I implemented the feature properly (yes, that's the second try), and will provide a PR soon.

PR #632, opened 2016-12-17

TheAssassin:
This commit introduces the Red Eclipse URL scheme that can be used to start Red Eclipse and immediately connect to a Red Eclipse server without having to type a connect command or using the server browser to find a specific server.

redeclipse://hostname:port
redeclipse://hostname

Slashes appended to the URL are ignored, password and port arguments are optional.

The parser uses the same functionality as the /connect command, which means it provides similar behavior.

Beware you need to configure your system to launch the Red Eclipse client binary when redeclipse:// URLs are opened (e.g. in a Browser or a desktop shortcut). You can find an example for Linux desktops in the community repository (https://github.com/red-eclipse/community).

**This is the second attempt on this feature**. See https://github.com/red-eclipse/base/pull/631 for  information on the first implementation.

### DIFF
--- a/src/engine/main.cpp
+++ b/src/engine/main.cpp
@@ -924,6 +924,17 @@ int main(int argc, char **argv)
 
     char *initscript = NULL;
     initing = INIT_RESET;
+
+    // used to support redeclipse:// URIs
+    // see below for more information
+    const char reprotoprefix[] = "redeclipse://";
+    const int reprotooffset = strlen(reprotoprefix);
+    char* connectpassword = NULL;
+    char* connecthost = NULL;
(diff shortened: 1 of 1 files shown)


TheAssassin:
I'm working out a solution that contains the requested changes collaboratively. Thus, I'm going to create a third PR about this feature.

Croydon:
@TheAssassin Maybe it may have been nicer if you just delete your old commits/reset your branch and push new commits to your branch. Then you wouldn't need new pull requests every time :)

TheAssassin:
Thanks for the tip. I basically did that, but didn't know it'd reset my PR like this then.

PR #633, opened 2016-12-17

TheAssassin:
This commit introduces the Red Eclipse URL scheme that can be used to start Red Eclipse and immediately connect to a Red Eclipse server without having to type a connect command or using the server browser to find a specific server.

redeclipse://hostname:port
redeclipse://hostname

Slashes appended to the URL are ignored, password and port arguments are optional. Spaces are allowed after the prefix (...://), but they are only useful in the password part.

The parser uses the same functionality as the /connect command, which means it provides similar behavior.

Beware you need to configure your system to launch the Red Eclipse client binary when redeclipse:// URLs are opened (e.g. in a Browser or a desktop shortcut). You can find an example for Linux desktops in the community repository (https://github.com/red-eclipse/community).

This is the third attempt to get this merged. See PRs https://github.com/red-eclipse/base/pull/631 and https://github.com/red-eclipse/base/pull/632 for more information on the other ones.

### DIFF
--- a/src/engine/main.cpp
+++ b/src/engine/main.cpp
@@ -1108,8 +1108,11 @@ int main(int argc, char **argv)
     localconnect(false);
     resetfps();
 
-    if(connecthost && *connecthost) connectserv(connecthost, connectport, connectpassword);
-    else conoutf("\frmalformed commandline argument: %s", reprotoarg);
+    if(reprotoarg)
+    {
+        if(connecthost && *connecthost) connectserv(connecthost, connectport, connectpassword);
+        else conoutf("\frmalformed commandline argument: %s", reprotoarg);
+    }
 
     // housekeeping
     if(connectstr)
--- a/src/engine/main.cpp
+++ b/src/engine/main.cpp
@@ -925,8 +925,12 @@ int main(int argc, char **argv)
     char *initscript = NULL;
     initing = INIT_RESET;
 
-    // used to support redeclipse:// URIs
-    // see below for more information
+    // redeclipse:// URI support
+    // examples:
+    // redeclipse://password@hostname:port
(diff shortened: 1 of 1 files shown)


src/engine/main.cpp
```diff
@@ -961,6 +977,44 @@ int main(int argc, char **argv)
                 if(!serveroption(argv[i])) gameargs.add(argv[i]);
                 break;
         }
+
```
qreeves: I'm not a big fan of comments between connected blocks like if/else, it's safe enough to move to inside the statement at the top. Also, is it necessary to to repeat the `redeclipse:// URI support` header? I don't mind comments, it's just that's a lot of new lines to scroll through.

src/engine/main.cpp
```diff
@@ -1056,6 +1110,22 @@ int main(int argc, char **argv)
...
+    if(connecthost && *connecthost) connectserv(connecthost, connectport, connectpassword);
+    else conoutf("\frmalformed commandline argument: %s", reprotoarg);
+
+    // housekeeping
+    if(connectstr)
+    {
+        delete[] connectstr;
+        connectstr = NULL;
+    }
+    if(reprotoarg) {
```
qreeves: New line before `{`

src/engine/main.cpp
TheAssassin: Yup, if you look where the `connect*` variables are set, it's easy to understand what that code snippet belongs to.
The other comment could be moved to the variable creation part.

acerspyro:
Someone try and abuse this, maybe by spamming it in JS?

TheAssassin:
Like with any other system that supports custom protocols (e.g. Mumble, ...).

PR #638, opened 2016-12-19
Prevent moderators from quarantining equally or higher privileged players (closes #637)

### LINKED ISSUE #637: Prevent moderators from quarantining other mods
In cooperation with @acerspyro, I discovered that moderators (or better) can put other moderators (or better) into quarantine, which should be forbidden to prevent trolling.
We should implement a privilege level check that checks the target's privilege and prevents the operation if they have the same or higher privileges.
This is related to #636.

TheAssassin:


src/game/server.cpp
```diff
@@ -6945,6 +6945,11 @@ namespace server
                         break;
                     }
                     bool spec = val != 0, quarantine = cp != ci && val == 2, wasq = cp->quarantine;
+                    if(quarantine && ci->privilege <= cp->privilege)
+                    {
+                        srvmsgf(ci->clientnum, "\frerror: you may not quarantine higher or equally privileged player %s", colourname(cp));
+                        break;
+                    }
```
qreeves: Direct comparisons between client privileges will be skewed unless you mask out the high bits that determine local/global, etc. To get the correct representation you use `(ci->privilege&PRIV_TYPE)`. Also, most privilege comparisons spit out the text (red)"access denied, ..."

PR #641, opened 2016-12-20
Fixed french accents? How does that even break?

acerspyro:


qreeves:
Happens when someone saves with the ANSI instead of UTF-8 encoding.

acerspyro:
@qreeves Gotcha, although I used GitHub's editor, which handles them just fine. So I assume a previous edit broke them.

PR #644, opened 2016-12-21
Include directory enet/include in CMakeLists.txt

Robalni:
cube.h tries to include enet.h at line 61 but that gives a compiler error because the compiler doesn't look in the right directory.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -90,6 +90,7 @@ add_subdirectory(enet)
 
 # configure local includes
 include_directories(
+    ${CMAKE_CURRENT_SOURCE_DIR}/enet/include
     ${CMAKE_CURRENT_SOURCE_DIR}/engine
     ${CMAKE_CURRENT_SOURCE_DIR}/game
     ${CMAKE_CURRENT_SOURCE_DIR}/include


TheAssassin:
The CMake configuration works fine on my computer actually, I don't know why it does not on yours. I use CMake 3.5.something. What's your version?

Robalni:
I guess ${ENET_INCLUDE_DIRS} is supposed to be this path but enet doesn't set that variable. I checked it with message() in the top-level CMakeLists and it is empty.

I use cmake 3.0.2

TheAssassin:
I doubt that there should be such a variable in `src/CMakeLists.txt` after `add_subdirectory(enet)`.
I can confirm that the includes are not modified globally. But according to the CMake docs, this is not intented anyway.
The directory is included in enet's `CMakeLists.txt`, and is then added to any target that uses the library. Since we call `target_link_libraries(redeclipse_client${BIN_SUFFIX} enet)` etc., the include is added to these targets. This works fine on my Ubuntu Xenial with CMake 3.5.2.

I will have to further investigate this, but hardcoding the path in the root `CMakeLists.txt` is not the way to go IMO.
Perhaps we need to increase the CMake version. What OS do you use (including version)?

Robalni:
One solution could be to use [target_include_directories](https://cmake.org/cmake/help/v3.2/command/target_include_directories.html) in enet's CMakeLists and specify PUBLIC as the scope.Then all targets that link to enet will get the right include directory.

The only problem with that is that genkey doesn't link to enet but it includes cube.h which includes enet/enet.h

TheAssassin:
Then we would have to hardcode the paths for that, too. Seems like there is no good way to populate the global includes from enet's file. Therefore, your initial solution appears acceptable to me.

PR #648, opened 2016-12-29
appdata: Use valid SPDX license IDs

kalev:
This fixes the license IDs in the appdata file to conform to the SPDX specification. Without this, appstream-builder that is used in Fedora rejects the appdata files.

### DIFF
--- a/src/install/nix/redeclipse.appdata.xml.am
+++ b/src/install/nix/redeclipse.appdata.xml.am
@@ -2,7 +2,7 @@
 <component type="desktop">
   <id>@APPNAME@.desktop</id>
   <metadata_license>CC-BY-SA-3.0+</metadata_license>
-  <project_license>ZLIB and CC-BY-SA-3.0+</project_license>
+  <project_license>Zlib and CC-BY-SA-3.0+</project_license>
   <description>
     <p>
       Red Eclipse is a fun-filled new take on the first person arena shooter,


TheAssassin:
If we remove the `+`, we change the license information to a wrong value. It is apparently intended to say "use CC-By-SA-3.0 or any later version of the license", the change means "you must use CC-By-SA-3.0 forever".
Is there a way to represent this in the metadata? Otherwise we will have to discuss this further, and thus have to reject the PR.

kalev:
Ohhhhh, you are right! I totally missed that SPDX now supports "+" at the end of license identifiers in their new version. I'll go fix appstream-builder instead to correctly handle this. Thanks, and sorry for the confusion :)

kalev:
Reopening for a remaining ZLIB -> Zlib change; the + is now handled correctly on the appstream-builder side with https://github.com/hughsie/appstream-glib/pull/143 but the all-caps ZLIB still makes it error out.

kalev:
I've pushed a fixed up commit for that.

TheAssassin:
I'm not into Fedora packaging, can you cite some source that shows what licenses Fedora accepts except the projects you contribute to? I'd just like to see that this does not break Fedora packaging, but I don't have time to create a Docker container atm.

kalev:
https://people.freedesktop.org/~hughsient/appdata/ which is linked from https://fedoraproject.org/wiki/Packaging:AppData

The remaining issue here is that "ZLIB" isn't a valid license identifier as per the SPDX license list (https://spdx.org/licenses/), instead it has to read "Zlib". This is what the patch addresses.

TheAssassin:
Okay, after reading a bit into the topic, this looks good to me. Thanks for the patch!

kalev:
Thanks!

PR #649, opened 2016-12-31
Make player preview rotation controllable

### LINKED ISSUE #44: rotation of player preview
With even more vanity items to explore, it would be cool to have some control of the player preview's rotation (at least for the larger one, "playerprev"). Pausing the rotation does already help much to look at vanity items, so what about a more convenient way than messing with /gamespeed or /gamepaused? An interactive way to rotate the preview model would be a very nice extra.

Robalni:
You can control it by hovering with the mouse cursor and when you leave it, it continues normal rotation.

I think this fixes #44

### DIFF
--- a/src/engine/rendergl.cpp
+++ b/src/engine/rendergl.cpp
@@ -1606,9 +1606,10 @@ namespace modelpreview
     }
 }
 
-vec calcmodelpreviewpos(const vec &radius, float &yaw)
+vec calcmodelpreviewpos(const vec &radius, float &yaw, float setyaw)
 {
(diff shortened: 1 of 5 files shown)


qreeves:
The effect here is really jarring and implemented in a very static sort of way that allows for too many accidental interactions. Ideally this would be click and drag on either axis, possibly with some sliders along each edge to control it. Switching it from spinning and user controlled shouldn't snap between them but rather store the current spin and start rotating from there.

Robalni:
I don't think I'm going to improve this right now so I'm closing it.

qreeves:
PR is superseded by the nnUI port anyway, I'll look at implementing this in the new system.

PR #655, opened 2017-01-15
Fix infinite impulse glitch

zaquest:
Related forum topic http://redeclipse.net/forum/viewtopic.php?f=7&t=999&p=9127
Helps to improve time on absorption by a lot. Makes camping by infinite wallrunning possible. SniperGoth, and people in the forum topic think this should be fixed and is a potential game breaker.


qreeves:
I'm not sure of the repercussions with regard to transitioning to a new impulse state without fully resetting the player's parkour state, but I guess we'll see.

Issue #656, opened 2017-01-18
Limit length of custom version suffix


TheAssassin:
At the moment, it is possible to set the `RE_BRANCH` environment variable very long strings, which increases the scoreboard width a lot. We should limit its length in some way.
A hard coded limit should do it, there is no need for a variable. We have to sanitize user input, so if the string is longer than e.g. 20 characters, we'll just clip it.
Limiting it to a set of predefined strings was suggested too, but this is more effective. It could be considered a feature to be able to set version strings to custom values, too.

qreeves:
Yeah, I considered it a feature to set them as well. Probably other things that need this as well.

PR #657, opened 2017-01-18
Limit the sent branch length to 16 characters.

shacknetisp:
This will cut off branch text longer than `MAXBRANCHLEN` so that clients will not try to display all of it as described in #656.


qreeves:
Yeah, but if you're going to truncate the string it needs to be done everywhere a branch string gets set, this includes `versionbranch` and `game::start()`, otherwise the client will still see the non-truncated string everywhere in their local client.

shacknetisp:
I actually did consider this but decided not to since there might be a case where a very long version string is actually useful to the client, but I'll add the truncations everywhere now.

acz13:
>there might be a case where a very long version string is actually useful to the client

Example of such a case?

shacknetisp:
I can't think of any specific examples of why a client would be using long branch strings for identification, that's why I implemented this commit.

PR #659, opened 2017-01-23
Add DISC_AUTH for when clients authenticate with the same handle.

shacknetisp:
This replaces the old DISC_KICK method in this situation, clarifying what is happening to those reading the message.


TheAssassin:
Good idea, improves the auth system UX a lot.

PR #661, opened 2017-01-25
Improve IRC and log privacy

TheAssassin:
Until now, a user's system's username and hostname have been written to the
log files and were used as identification for the in-game IRC client.

Both is a privacy issue. Being in the log, if an unexperienced user pastes
their log on the internet, both information is shared with third parties.
Using it for the IRC client is even more dangerous, because if some user
accidentally pressed I ingame, the username would be leaked automatically.
As the user is not even informed about this, it might even be illegal in
several countries.

As a result, systemuser and systemhost variables have been removed as they
would no longer be used in the codebase.

The issue has previously been discussed on the forums more than a year ago
(http://redeclipse.net/forum/viewtopic.php?f=7&t=814), but resulted in some
"might be useful". Apparently, this is not an excuse for creating potential
privacy issues.

### DIFF
--- a/src/engine/irc.cpp
+++ b/src/engine/irc.cpp
@@ -920,7 +920,7 @@ void ircslice()
                     if(*n->passkey) ircsend(n, "PASS %s", n->passkey);
                     copystring(n->nick, n->mnick);
                     ircsend(n, "NICK %s", n->mnick);
-                    if(!*n->ident) copystring(n->ident, systemuser);
+                    if(!*n->ident) copystring(n->ident, VERSION_UNAME);
                     ircsend(n, "USER %s +iw %s :%s v%s-%s%d-%s (%s)", n->ident, n->ident, versionname, versionstring, versionplatname, versionarch, versionbranch, versionrelease);
                     n->lastnick = clocktime;
                     n->state = IRC_CONN;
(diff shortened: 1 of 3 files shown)


src/engine/irc.cpp
```diff
@@ -920,7 +920,7 @@ void ircslice()
                     if(*n->passkey) ircsend(n, "PASS %s", n->passkey);
                     copystring(n->nick, n->mnick);
                     ircsend(n, "NICK %s", n->mnick);
-                    if(!*n->ident) copystring(n->ident, systemuser);
+                    if(!*n->ident) copystring(n->ident, VERSION_UNAME);
```
qreeves: `versionuname`

src/engine/server.cpp
```diff
@@ -50,7 +50,6 @@ VAR(IDF_READONLY, versionisserver, 0, 0, 1);
 #endif
 ICOMMAND(0, platname, "ii", (int *p, int *g), result(*p >= 0 && *p < MAX_PLATFORMS ? (*g!=0 ? plat_longname(*p) : plat_name(*p)) : ""));
 
-SVAR(IDF_READONLY, systemuser, "none");
 SVAR(IDF_READONLY, systemhost, "unknown");
```
qreeves: Left unused variable.

PR #666, opened 2017-01-29
OpenBSD/rm command line does not support verbose flag

devnexen:
... as the port uses the old Makefile way, it is useful :-)


acz13:
Is this only OpenBSD?

devnexen:
FreeBSD supports it

TheAssassin:
I checked with the man entries and yeah, OpenBSD's `rm` does not support `-v`. I didn't check for other BSD flavors, as this PR is explicitly fixing the issue for OpenBSD.
We can always add other BSDs or Linux distros later.

qreeves:
Thanks for implementing this in a portable way!

PR #670, opened 2017-02-12
Fix wrong desktop entry version

IceflowRE:
The desktop file needs to follow at least version 1.1. Since the "keywords" key was added in 1.1: https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#idm140271348128080
And because 1.2 does not add something, this file is not following, we can use the 1.2 version.

### DIFF
--- a/src/install/nix/redeclipse.desktop.am
+++ b/src/install/nix/redeclipse.desktop.am
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Type=Application
-Version=1.0
+Version=1.2
 Name=Red Eclipse
 GenericName=First-person shooter game
 GenericName[es]=Juego de tiros en primera persona


TheAssassin:
I don't see a reason to use 1.2 if we could just use 1.1, too. Wouldn't using 1.2 break compatibility with older systems?

IceflowRE:
Using a version which does not exist/ cant be used has no effect then.

TheAssassin:
Well I guess most desktop file parsers would also just try to parse newer versions, anyway (like HTML parsers who try to make the best of what they get). And 1.2 seems to be around for a while (couldn't find any "release notes", but the search wasn't more than the first page of Google results). I guess this can be safely merged.

PR #672, opened 2017-02-18
Add last hit gladiator score.

shacknetisp:
This provides a bonus score (like a frag by default) to the client who last hit a falling client in gladiator before their death. This will make it easier to earn points in gladiator without survivor or duel, meaning that players who have just joined will not have an advantage. The time is longer than normal assists since some maps or launch angles can have a long fall time.


src/game/server.cpp
```diff
@@ -4097,15 +4097,33 @@ namespace server
...
             {
                 log.add(assist->clientnum);
-                if(points) givepoints(assist, points, m_points(gamemode, mutators), true);
+                if(points)
+                {
+                    if(!lastpoints || i != last)
+                    {
+                        if(millis-m->damagelog[i].millis <= G(assistkilldelay)) givepoints(assist, points, m_points(gamemode, mutators), true);
+                    }
+                    else
```
qreeves: This doesn't need braces.

src/game/server.cpp
```diff
@@ -4097,15 +4097,33 @@ namespace server
...
-    void gethistory(clientinfo *m, clientinfo *v, int millis, vector<int> &log, bool clear = false, int points = 0)
+    void gethistory(clientinfo *m, clientinfo *v, int millis, vector<int> &log, bool clear = false, int points = 0, int lastpoints = 0)
     {
-        loopv(m->damagelog) if(m->damagelog[i].clientnum != v->clientnum && millis-m->damagelog[i].millis <= G(assistkilldelay))
+        int last = -1;
+        if(lastpoints)
+        {
+            loopv(m->damagelog) if(m->damagelog[i].clientnum != v->clientnum && millis-m->damagelog[i].millis <= G(lasthitdelay))
+            {
+                if(last == -1 || m->damagelog[i].millis > m->damagelog[last].millis) last = i;
```
qreeves: We don't usually explicitly check for -1 unless we're after that as a value.

Alternatives: `m->damagelog.inrange(last)` or `last < 0`

src/game/server.cpp
shacknetisp: Without these braces the else statement below is interpreted as part of
`if(millis-m->damagelog[i].millis <= G(assistkilldelay))`

PR #674, opened 2017-02-20
Add a variable to toggle MOUSE3 escaping the current UI.

shacknetisp:
Variable name and usage description might need some tweaking, let me know if they do.


qreeves:
This is a bit hacky. I had plans to make a `uibind` system already, so I might just incorporate the hard coded stuff into that as well.

shacknetisp:
Ok, this was just for if the binds would remain hard coded.

PR #675, opened 2017-02-21
Do not display demo errors when just loading the maps cache.

shacknetisp:
Several somewhat spammy errors currently appear in the console during cache filling when there are incompatible demos found, this quiets those errors since incompatibility should be indicated in the selection ui.


qreeves:
Actually, it should just send the message to `CON_MESG` which isn't shown unless you have `fullconsole` on or the command prompt open. Once the UI port is complete we won't even have a message console on the screen by default, it will be replaced by a debug console, with separate things for chat and game events (like other games).

PR #677, opened 2017-02-21
Automatically build AppImages on Semaphore

TheAssassin:
The PR introduces auto builds for bleeding-edge AppImages that can be downloaded by users who want to try out the latest features. It uses https://github.com/TheAssassin/redeclipse-appimage and a Docker container to build the AppImage in a Debian wheezy container.
The only feature missing is some kind of deployment to a server where users can download it. I guess @qreeves has to add that though, as I don't know much about his infrastructure.
I can also provide some space on my server for that.


probonopd:
Could even deploy to GitHub Releases using something like [this](https://github.com/probonopd/uploadtool).

PR #678, opened 2017-02-25
Use CMAKE_BUILD_TYPE to set compiler flags

Robalni:
This makes it possible to build with debug information and without optimization using this command:

```
cmake -DCMAKE_BUILD_TYPE=debug
```

If no build type is specified then release is used.


TheAssassin:
It's a good idea to implement this anyway, so thanks for this @Robalni.

PR #679, opened 2017-03-04
Added health percentage above the heads of players near their names

echo-devim:
The percentage changes its colour based on its level

### DIFF
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -3169,21 +3169,22 @@ namespace game
         float blend = aboveheadblend*trans;
         if(aboveheadnames && d != player1)
         {
-	        const char *name = colourname(d);
-	        int health_length;
-	        int h = d->health;
-	        for (health_length=0; h!=0; health_length++, h=h/10){}
-	        health_length += 17; /* Fixed amount of extra-characters manually calculated */
-	        char name_health[strlen(name)+health_length];
-	        strcpy(name_health, name);
-	        if (aboveheadhealth && (d->health > 0)) {
-		        char health[health_length];
-		        int health_colour = 0x45db1c; /* green */
-		        if (d->health < 30)
-		            health_colour = 0xe03a1d; /* red */
-	            else if (d->health < 70)
-		            health_colour = 0xe09f1d; /* yellow */	
-		        sprintf(health, "\fs\f[%d]%d%%\fS", health_colour, d->health);
+	    const char *name = colourname(d);
+	    int health_length;
+	    int h = d->health;
+	    for (health_length=0; h!=0; health_length++, h=h/10);
+	    health_length += 17;
+	    char name_health[MAXNAMELEN+health_length];
+	    strcpy(name_health, name);
+	    if (aboveheadhealth && (d->health > 0))
+	    {
+	        char health[health_length];
+		int health_colour = 0x45db1c; /* green */
+		if (d->health < 30)
+		    health_colour = 0xe03a1d; /* red */
+                else if (d->health < 70)
+		    health_colour = 0xe09f1d; /* yellow */	
+		    sprintf(health, "\fs\f[%d]%d%%\fS", health_colour, d->health);
 	            strcat(name_health, health);
             }
             pos.z += aboveheadnamessize/2;
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -3169,18 +3169,22 @@ namespace game
         float blend = aboveheadblend*trans;
         if(aboveheadnames && d != player1)
         {
-	    const char *name = colourname(d);
-	    char name_health[strlen(name)+100];
-	    strcpy(name_health, name);
-	    if (aboveheadhealth && (d->health > 0)) {
-		char health[100];
-		int health_colour = 0x45db1c; /* green */
-		if (d->health < 30)
-		    health_colour = 0xe03a1d; /* red */
-	        else if (d->health < 70)
-		    health_colour = 0xe09f1d; /* yellow */	
-		sprintf(health, "\fs\f[%d]%d%%\fS", health_colour, d->health);
-	        strcat(name_health, health);
+	        const char *name = colourname(d);
+	        int health_length;
+	        int h = d->health;
+	        for (health_length=0; h!=0; health_length++, h=h/10){}
+	        health_length += 17; /* Fixed amount of extra-characters manually calculated */
+	        char name_health[strlen(name)+health_length];
+	        strcpy(name_health, name);
+	        if (aboveheadhealth && (d->health > 0)) {
+		        char health[health_length];
+		        int health_colour = 0x45db1c; /* green */
+		        if (d->health < 30)
+		            health_colour = 0xe03a1d; /* red */
+	            else if (d->health < 70)
+		            health_colour = 0xe09f1d; /* yellow */	
+		        sprintf(health, "\fs\f[%d]%d%%\fS", health_colour, d->health);
+	            strcat(name_health, health);
             }
             pos.z += aboveheadnamessize/2;
             part_textcopy(pos, name_health, PART_TEXT, 1, 0xFFFFFF, aboveheadnamessize, blend*aboveheadnamesblend);
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -3170,10 +3170,7 @@ namespace game
         if(aboveheadnames && d != player1)
         {
 	    const char *name = colourname(d);
-	    int health_length;
-	    int h = d->health;
-	    for (health_length=0; h!=0; health_length++, h=h/10);
-	    health_length += 17;
+	    int health_length = log10(INT_MAX) + 18;
 	    char name_health[MAXNAMELEN+health_length];
 	    strcpy(name_health, name);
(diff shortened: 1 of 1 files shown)


src/game/game.cpp
```diff
@@ -25,7 +25,7 @@ namespace game
         player1->version.gpuglslver = glslversion;
         player1->version.crc = versioncrc;
         if(player1->version.branch) delete[] player1->version.branch;
-        player1->version.branch = newstring(versionbranch, MAXBRANCHLEN);
+        player1->version.branch = newstring(versionbranch);
```
TheAssassin: This line of code was correct before, I am not sure why you changed it for this PR. Especially since it does not have anything to do with the PR's purpose. Please undo this change.

Calinou:
I have a few concerns with this pull request:

Most `abovehead*` variables allow you to toggle between displaying information only above teammates, only above enemies, or above both. This one doesn't.
Coloring uses hard-coded discrete steps, it should use continuous coloring (using a simple mathematical formula).

It should also be discussed for balance reasons whether this should be allowed to be seen on enemies or not.

echo-devim:
Another thing, I forgot to use the right buffer size. If you look at the code I used a "large" buffer (health[100]) when it isn't needed.

src/game/game.cpp
```diff
@@ -3168,8 +3169,21 @@ namespace game
         float blend = aboveheadblend*trans;
         if(aboveheadnames && d != player1)
         {
+	    const char *name = colourname(d);
+	    char name_health[strlen(name)+100];
```
TheAssassin: The use of `100` as a constant looks rather random. It is rather easy to calculate how many additional characters are needed.

IceflowRE:
Is it only me or is the intendention wrong now?

src/game/game.cpp
```diff
@@ -3169,18 +3169,22 @@ namespace game
...
-		sprintf(health, "\fs\f[%d]%d%%\fS", health_colour, d->health);
-	        strcat(name_health, health);
+	        const char *name = colourname(d);
+	        int health_length;
+	        int h = d->health;
+	        for (health_length=0; h!=0; health_length++, h=h/10){}
+	        health_length += 17; /* Fixed amount of extra-characters manually calculated */
+	        char name_health[strlen(name)+health_length];
+	        strcpy(name_health, name);
+	        if (aboveheadhealth && (d->health > 0)) {
```
TheAssassin: Please put the brace on a new line.

src/game/game.cpp
```diff
@@ -3169,18 +3169,22 @@ namespace game
...
+		        char health[health_length];
+		        int health_colour = 0x45db1c; /* green */
+		        if (d->health < 30)
+		            health_colour = 0xe03a1d; /* red */
+	            else if (d->health < 70)
```
TheAssassin: Wrong indentation.

src/game/game.cpp
```diff
@@ -3169,18 +3169,22 @@ namespace game
...
-		    health_colour = 0xe03a1d; /* red */
-	        else if (d->health < 70)
-		    health_colour = 0xe09f1d; /* yellow */	
-		sprintf(health, "\fs\f[%d]%d%%\fS", health_colour, d->health);
-	        strcat(name_health, health);
+	        const char *name = colourname(d);
+	        int health_length;
+	        int h = d->health;
+	        for (health_length=0; h!=0; health_length++, h=h/10){}
+	        health_length += 17; /* Fixed amount of extra-characters manually calculated */
```
TheAssassin: There is no need for this inline comment. If you want a comment there, you should explain why this constant value is there.

src/game/game.cpp
```diff
@@ -3169,18 +3169,22 @@ namespace game
...
+	        char name_health[strlen(name)+health_length];
```
TheAssassin: Variable length char arrays don't seem to be the way to go here. You should use the `MAXNAMELEN` constant instead, or just use the `string` type instead.

Dziq84:
what ? above head - health display ? its an rpg game ? I fear this can lead to spoil game too much and even less noob friendly, like everyone will choose targets based on hp left (lets kill that one with less hp since its easier). damn, only I thought like that ?

TheAssassin:
No @Dziq84, we discussed (and are discussing) that on IRC. I already mentioned it might lead to even more rifle camping.

src/game/game.cpp
```diff
@@ -3170,7 +3171,7 @@ namespace game
         if(aboveheadnames && d != player1)
         {
 	    const char *name = colourname(d);
-	    int health_length = log10(INT_MAX) + 18;
+	    int health_length = MAXINTDIGITS + 18;
```
TheAssassin: Since this is a constant, you shouldn't calculate the sum on every run.

```
static const int health_length = MAXINTDIGITS + 18;
static const int name_health_length = MAXNAMELEN + health_length;

File: src/game/game.cpp [hunk 8]
```diff
@@ -3168,8 +3170,24 @@ namespace game
         float blend = aboveheadblend*trans;
         if(aboveheadnames && d != player1)
         {
+	    const char *name = colourname(d);
+	    static int health_length = MAXINTDIGITS + 18;
+            static const int name_health_length = MAXNAMELEN + health_length;
+	    char name_health[name_health_length];
```
shacknetisp: Replace this with the string type: `string name_health`

src/game/game.cpp
```diff
@@ -3168,8 +3170,24 @@ namespace game
...
+	    strcpy(name_health, name);
```
shacknetisp: Here you'd use the copystring function: `copystring(name_health, name)`

src/game/game.cpp
```diff
@@ -2,6 +2,7 @@
 #include "game.h"
 namespace game
 {
+    static const int MAXINTDIGITS = log10(INT_MAX);
```
shacknetisp: This line is not necessary when using the string type, which does not require a length specified.

src/game/game.cpp
```diff
@@ -3168,8 +3170,24 @@ namespace game
...
+	    if (aboveheadhealth && (d->health > 0))
+	    {
+	        char health[health_length];
+		int health_colour = 0x45db1c; /* green */
+		if (d->health < 30)
+		    health_colour = 0xe03a1d; /* red */
+                else if (d->health < 70)
+		    health_colour = 0xe09f1d; /* yellow */	
+		sprintf(health, "\fs\f[%d]%d%%\fS", health_colour, d->health);
+	        strcat(name_health, health);
```
shacknetisp: Replace these two lines with the `concformatstring` function: `concformatstring(name_health, "\fs\f[%d]%d%%\fS", health_colour, d->health)`

src/game/game.cpp
```diff
@@ -3168,8 +3170,24 @@ namespace game
         float blend = aboveheadblend*trans;
         if(aboveheadnames && d != player1)
         {
+	    const char *name = colourname(d);
+	    static int health_length = MAXINTDIGITS + 18;
+            static const int name_health_length = MAXNAMELEN + health_length;
```
shacknetisp: These two lines are not necessary when using the string type, which does not require a length specified.

src/game/game.cpp
```diff
@@ -3168,8 +3170,24 @@ namespace game
...
+	    char name_health[name_health_length];
+	    strcpy(name_health, name);
+	    if (aboveheadhealth && (d->health > 0))
+	    {
+	        char health[health_length];
```
shacknetisp: You do not need this variable with a direct concatenation below.

graphitemaster:
not only that formatting should be done with deffmtstring or what ever it's called

TheAssassin:
@graphitemaster `defformatstring(...)`.

qreeves:
The problem with the indentation is that this PR is using TAB characters whereas RE source code style uses four SPACE characters. TAB is normally eight characters long depending on how you set your viewer up, RE does not allow the use of TAB characters in the source code for indentation.

For colouring things based on an amount, there is `hud::skewcolour()`. For getting the spawn and maximum health values *for each different actor type* (bots have a different set of variables than human players) there are `m_spawnhealth()` and `m_maxhealth()` macros.

There's also an array of string functions we have in order to get around the limitations of the normal string functions and also be able to use our various string types. There's `conformatstring(dest, format, ...)` to get around needing to `sprintf()` with an intermediate variable which you'd then have to `strcat()`.

That being said, there are some concerns from the community about this topic regarding "giving away too much information", so I'd suggest the discussion be moved to the [forum topic](http://redeclipse.net/forum/viewtopic.php?f=5&t=1162) for this issue.

All in all, while I appreciate the interest in providing a feature implementation, this submission fails to follow the established style in many ways, and it's obvious you're still figuring out what everything does. Note that I'm also considering removing the whole "abovehead" system and re-implementing it screen-space rather than world-space (possibly be integrated with the new UI), so for now I think we'll close this PR.

I realise you're relatively new here as far as interacting with the development team, so please don't let the outcome of this PR discourage you from continuing to try in the future, you'll get the hang of the code in a while (I'm not a real programmer and I get by well enough with the help of everyone else). As always, I'll try to help those genuinely interested in contributing and attempt to point them in the right direction.

PR #681, opened 2017-03-20
Remove duplicate savedscore functions

zaquest:
Remove `findstatsscore` and add parameter to `findscore` to operate on both `savedscores` and `savedstatsscores` vectors. Remove `scorestatscmp` and update `scorecmp` to behave like `scorestatscmp` which provides better detection of whether it's the same player or not.

### DIFF
--- a/src/game/server.cpp
+++ b/src/game/server.cpp
@@ -2781,7 +2781,7 @@ namespace server
         return false;
     }
 
-    savedscore *findscore(vector<savedscore>& scores, clientinfo *ci, bool insert)
+    savedscore *findscore(vector<savedscore> &scores, clientinfo *ci, bool insert)
     {
         uint ip = getclientip(ci->clientnum);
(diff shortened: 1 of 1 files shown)


src/game/server.cpp
```diff
@@ -2777,18 +2777,11 @@ namespace server
...
 
-    bool statsscorecmp(clientinfo *ci, uint ip, const char *name, const char *handle, uint clientip)
-    {
-        if(ci->handle[0] && !strcmp(handle, ci->handle)) return true;
-        else if(!ci->handle[0] && ip && clientip == ip && !strcmp(name, ci->name)) return true;
-        return false;
-    }
-
-    savedscore *findscore(clientinfo *ci, bool insert)
+    savedscore *findscore(vector<savedscore>& scores, clientinfo *ci, bool insert)
```
qreeves: In this codebase, the `&` is written as part of the value name, not the type.

`vector<savedscore> &scores`

PR #684, opened 2017-04-22
Use temp directory instead of maps directory for sent CRC-specified maps.

### LINKED ISSUE #626: Maps with CRC are not openable
If you try to start editing map that you've downloaded from the server (which has the name "mapname_0x12345678.mpz"), you get an empty map.

This bug only shows when you don't have the map with the same name but without CRC.

shacknetisp:
This PR changes the map downloading system to use the `temp/` directory for storing CRC-downloaded map files, freeing `maps/` for solely `savemap`ed maps and fixing #626.

## Potential Additions

* Automatic cleaning of old files after some time, perhaps just through their modification timestamp.

Fixes #626

### DIFF
--- a/src/engine/worldio.cpp
+++ b/src/engine/worldio.cpp
@@ -95,7 +95,11 @@ void setnames(const char *fname, int type, int crc)
 
     if(strpbrk(fn, "/\\")) copystring(mn, fn);
     else formatstring(mn, "%s/%s", mapdirs[maptype].name, fn);
-    setsvar("mapname", mn);
+
+    // Any map with a CRC must be placed in temp/, so we add the prefix if it does not already exist.
+    if(crc != 0 && strstr(mn, "temp/") != mn && strstr(mn, "temp\\") != mn) prependstring(mn, "temp/");
+    // Set the mapname variable. If there is no CRC, we need to remove the temp/ prefix.
+    setsvar("mapname", (crc == 0 && strstr(mn, "temp/") == mn) ? (mn + 5) : mn);
 
     formatstring(mf, "%s%s", mapname, mapexts[maptype].name);
     setsvar("mapfile", mf);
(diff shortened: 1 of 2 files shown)


swooboo:
This is cool. Will this make it harder for people to suggest maps they created / modded?

shacknetisp:
No, it just modifies the system so that the `mapname_crc` maps are no longer kept in the main maps directory. Custom maps should be `/savemap [<name>]` as always, and they'll be easily loadable with no extra clutter.

The one situation where it could be confusing is if someone loads a received map that is stored in the temporary directory, and then they use `savemap`. The map would not be available to them directly in that case, but it is little different then when they would use it to save  `mapname_crc` which do not load and transfer properly anyway. This could perhaps be avoided by stripping away `temp/` and the CRC code from the `savemap` parameter.

src/engine/worldio.cpp
```diff
@@ -49,14 +49,38 @@ SVARF(IDF_WORLD, maptitle, "", fixmaptitle());
...
+    if(*r == '/' && r > name) return r;
+    else return NULL;
+}
+
 void setnames(const char *fname, int type, int crc)
 {
     maptype = type >= 0 || type <= MAP_MAX-1 ? type : MAP_MAPZ;
 
-    string fn, mn, mf;
+    string fn, mn, fnm, mf, cn = "";
```
IceflowRE: Can you add a comment, which says what the variable names means? :)

shacknetisp: The problem is I don't know what most of them mean!
I added `fnm` and `cn` myself though:

* `fnm`: "Full Name"
* `cn`: "CRC to be appended to name"

Not sure about adding a comment since there's no others around.

qreeves: It doesn't really matter when it comes to local variables, they often have abstract names for the sake of being compact and easy to type. For what it's worth, it's roughly: `fn` - file name, `mn` - map name, `mf` - map file.

IceflowRE: Good variable names matters everytime, it is torture to guess what it means, with documented names (or better names) its much much easier to understand and read the code. And a comment does not hurt?

qreeves:
Can you resolve the current conflicts so I can merge this?

IceflowRE:
Thanks for the comments!

PR #687, opened 2017-05-12
Update captureresetfloor to proper default value (1).

JeffBobbo:
Currently the default is set to 2, but this means that if I toss my enemies' flag, it resets to my local position. The original post I made addressing this wasn't particularly clear (and was mostly about discussion), but my intention was that wouldn't be the case, as I could potentially toss near death to make it harder for the enemy team to get it back again (but, if a team mate is in the right location, much easier for them).

Also added a comment explaining what the values actually are.

### DIFF
--- a/src/game/vars.h
+++ b/src/game/vars.h
@@ -304,7 +304,7 @@ GVAR(IDF_GAMEMOD, teambalancehighest, 0, 1, 1); // when moving players, move hig
 GVAR(IDF_GAMEMOD, racegauntletwinner, 0, 1, 1); // declare the winner when the final team exceeds best score
 
 GVAR(IDF_GAMEMOD, capturelimit, 0, 0, VAR_MAX); // finish when score is this or more
-GVAR(IDF_GAMEMOD, captureresetfloor, 0, 2, 1); // if tossed, reset to last floor pos. 1 = team tosses own flag, 2 = any flag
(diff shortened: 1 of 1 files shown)


src/game/vars.h
```diff
@@ -304,7 +304,7 @@ GVAR(IDF_GAMEMOD, teambalancehighest, 0, 1, 1); // when moving players, move hig
 GVAR(IDF_GAMEMOD, racegauntletwinner, 0, 1, 1); // declare the winner when the final team exceeds best score
 
 GVAR(IDF_GAMEMOD, capturelimit, 0, 0, VAR_MAX); // finish when score is this or more
-GVAR(IDF_GAMEMOD, captureresetfloor, 0, 2, 2); // if tossed, reset to last floor pos
+GVAR(IDF_GAMEMOD, captureresetfloor, 0, 2, 1); // if tossed, reset to last floor pos. 1 = team tosses own flag, 2 = any flag
```
qreeves: The values are "min", "default", "max", so the middle value is the one you should change, not the last one.

I think @shacknetisp made the "2" value, so it would probably be good to get his input on this PR.

shacknetisp:
The `2` default value was set in commit 499376cf10289b2a07e103b4395a1a50f4ca3a39 and originally added from #596, I haven't worked on anything about this yet. I do like this proposed default's functionality though.

src/game/vars.h
JeffBobbo: Fixed, my apologies.

TheAssassin:
Looks good now, can be [merged](http://devhumor.com/media/git-merge).

PR #688, opened 2017-05-12
Only allow whispers directly toward bots to affect them.

shacknetisp:
This fixes the issue where if you send a message to any actor, all actors on that host (both the human and his bots) will receive the whisper. The result is several bot whispers in reply if you want to contact the human or a single bot, this pull request has bots ignore whispers that are not aimed directly at them.

### DIFF
--- a/src/game/ai.cpp
+++ b/src/game/ai.cpp
@@ -1824,7 +1824,9 @@ namespace ai
                 }
             }
             else aimed = true;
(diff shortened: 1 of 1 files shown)


TheAssassin:
I was just about to write an issue about this. Looks good to me.

qreeves:
I'll need to review this one properly when I get a chance, as `aimed` is more about grabbing instances of broadcast messages prefixed with the names of the bot (or "bots"). If it is a whisper with a specific target that isn't a bot, you could probably escape out of this at the `return` statement at the beginning of this `scanchat` function.

PR #690, opened 2017-05-13
Fix random weapon selection. Close #689

### LINKED ISSUE #689: Random weapon selection isn't really random
So I tend to use the random weapon feature on my secondary slot, and use the filter to limit it to the sword and the shotgun.

Unfortunately, in >80% of all spawns (rough estimation), my secondary is the sword, and only very rarely I spawn with a shotgun instead. This is quite a problem, as it is quite annoying and... kinda... ruins the feature.

We should thus check how the next weapon is chosen. Ideally, a random variable with uniform distribution would be used for this. This means we should probably implement our own algorithm for randomness, as the features provided by C+(++) (without the standard library, to be fair) are very limited (they might be "random", but for our use case, we need something "better").

(Side note: C++11 contains facilities for this, see e.g. [cppreference.com](http://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution). I could test that at work, and it was really good at what it promised to do. Unfortunately, some people don't like `std::` (or stuff like `boost::`, where this is implemented, too). 

I'll try to come up with something independent, using the same algorithms.

zaquest:


src/game/game.h
```diff
@@ -892,40 +892,26 @@ struct clientstate
...
-            loopj(W_LOADOUT) if(canrandweap(j + W_OFFSET)) musthave--;
-            vector<bool> forced;
-            loopj(W_LOADOUT) forced.add(false);
             vector<int> aweap;
-            loopj(W_LOADOUT) aweap.add(loadweap.inrange(j) ? loadweap[j] : 0);
-            loopvj(aweap)
+            loopj(AA(actortype, maxcarry)) aweap.add(loadweap.inrange(j) ? loadweap[j] : 0);
+            vector<int> rand, forcerand;
+            for(int t = W_OFFSET; t < W_ITEM; t++)
+                if(!hasweap(t, sweap) && m_check(W(t, modes), W(t, muts), gamemode, mutators) && !W(t, disabled) && aweap.find(t) == -1)
```
qreeves: I prefer `aweap.find(t) < 0`

src/game/game.h
```diff
@@ -892,40 +892,26 @@ struct clientstate
...
+                    (canrandweap(t) ? rand : forcerand).add(t);
+            loopj(AA(actortype, maxcarry))
             {
                 if(!aweap[j]) // specifically asking for random
                 {
-                    for(int t = rnd(W_ITEM-W_OFFSET)+W_OFFSET, r = 0; r < W_LOADOUT; r++)
+                    vector<int> &randsrc = rand.length() > 0 ? rand : forcerand;
+                    if(randsrc.length() > 0)
```
qreeves: Maybe also `(!randsrc.empty())`

src/game/game.h
```diff
@@ -892,40 +892,26 @@ struct clientstate
...
-            loopj(W_LOADOUT) aweap.add(loadweap.inrange(j) ? loadweap[j] : 0);
-            loopvj(aweap)
+            loopj(AA(actortype, maxcarry)) aweap.add(loadweap.inrange(j) ? loadweap[j] : 0);
+            vector<int> rand, forcerand;
+            for(int t = W_OFFSET; t < W_ITEM; t++)
+                if(!hasweap(t, sweap) && m_check(W(t, modes), W(t, muts), gamemode, mutators) && !W(t, disabled) && aweap.find(t) == -1)
+                    (canrandweap(t) ? rand : forcerand).add(t);
+            loopj(AA(actortype, maxcarry))
             {
                 if(!aweap[j]) // specifically asking for random
```
qreeves: I didn't notice this before, but it should probably be `(aweap[j] <= 0)`

PR #696, opened 2017-06-13
Remove uninstall shortcut from Start, changed installer header image

acerspyro:
I think putting the uninstaller shortcut in the Start Menu contributes to clutter, as it forces the Red Eclipse shortcut to be inside of a folder, which is far from ideal in the case of a Start menu, especially on Windows 8, 8.1 and 10. I updated the installer script to not include the uninstaller in the Start Menu, and pull the Red Eclipse icon outside of its folder, in order to reduce the amount of clicks necessary to launch the game.

I also changed the installer's header picture from the ugly shaded red that was present before. I am open to suggestions on this one.


swooboo:
Where will it be? Will users be able to uninstall only with 'add or remove programs'?

TheAssassin:
Yes, of course. @acerspyro just found it annoying to be in the start menu. I am not quite sure why, I guess it's due to the fact that Windows 8+ pretty much rely on the search boxes and no longer on the directory tree. Since I'm not a Windows user, I can only guess here. @acerspyro, please elaborate on this a bit.

qreeves:
Doesn't really interfere with the search feature, all I can guess is that @acerspyro deems it clutter? (There's no description provided, tut tut)

TheAssassin:
What I meant is the uninstaller. Software developers have IIRC been recommended for a while to no longer put uninstallers in the start menu. Back in Win 7 times, when you were looking for some files using the start menu's search feature, you could be spammed with uninstallers. I could imagine the same thing happened with newer Windows releases.

acerspyro:
Right-clicking on the Red Eclipse icon in the Windows 10 Start Menu should present you with an "Uninstall" option.

Putting the shortcut in a folder prevents the icon from being seen at a glance.

The Start Menu is for launching applications, not removing them. Use the "Add/Remove Programs" control panel to do so. I have never seen someone head over to the Start Menu's application list to uninstall their software.

Moving the Red Eclipse icon outside of its folder reduces the amount of clicks needed in order to launch the game.

acerspyro:
Actually, after further investigation, it appears that the "Uninstall" option is missing from Red Eclipse's context menu in the Windows 10 Start Menu. Other NSIS applications seem to have the option, though. Maybe there's something I need to put in the NSIS script.

TheAssassin:
Next time, please send separate PRs, one per change, just like with commits. Avoid having to describe more than one feature in the title. And please write a description that tells us why you want to change it etc.

Conversations e.g. on IRC are useful, but they are not available to everyone, all the time. And we try to work transparently, which means that changes need to be visible to anyone.

Other than that, the NSI changes look good. But I'm not so sure about the header picture. Again, without a description, there's no way for me to see why this change was necessary. Please elaborate.

acerspyro:
This is what the installer image looks like:
image

PR #698, opened 2017-06-18
Making it possible to build server and genkey without SDL (when using CMake)

Robalni:
I tried to compile the Red Eclipse server on a server that did not have SDL installed but that failed because SDL is a required dependency for all targets in CMakeLists.txt.

This pull request makes SDL optional, and if it is not found then the client will not be compiled but the server and genkey will still be.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -16,7 +16,10 @@ find_package(PkgConfig REQUIRED)
 set(common_deps zlib)
 set(client_deps sdl2 SDL2_image SDL2_mixer gl)
 # will be set to false if the client dependencies are not installed
-set(build_client true)
+set(BUILD_CLIENT true CACHE BOOL "Build client?")
+if(NOT ${BUILD_CLIENT})
+    message(STATUS "Will not build client as requested")
+endif()
 
 # the client depends on almost all the source files
 file(GLOB client_sources engine/*.cpp game/*.cpp shared/*.cpp support/sqlite3.c)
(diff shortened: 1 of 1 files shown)


TheAssassin:
Very good idea, I didn't notice this unnecessary dependency before.

Though, I'm not a big fan of implicit configuration. They're just too automagic in my opinion. How about adding variables like `BUILD_CLIENT`, `BUILD_SERVER` and `BUILD_GENKEY` instead?

Robalni:
You mean so the user could do something like this?
```
cmake -DBUILD_CLIENT=false
```

TheAssassin:
Exactly. We could even add a warning that if it is set to true (default) and SDL and the other deps aren't found (and even exit at that state).

Something like "SDL could not be found. Client cannot be built. You can deactivate the client build by setting `-D BUILD_CLIENT=0`".

Robalni:
Do you think it should stop processing immediately when the client fails or continue processing CMakeLists.txt to possibly show other errors too?

TheAssassin:
I think it should exit immediately, then the user knows why. Users often don't scroll back but expect the error message to be displayed immediately.

TheAssassin:
I just noticed that this is going to behave differently now on different systems. The server still links against SDL2 due to the fact that `link_libraries` is used in the section that is guarded by `BUILD_CLIENT`. Thus, if the client build is enabled, it links to SDL2, if not, it does not. That means there will be two versions of the server binary, which is ambiguous and likely to cause errors when distributing them.

Instead, we need to move the definition of the binaries up, and then use `target_link_libraries` for the client target for that instead. Same goes for zlib, I'm pretty sure that `genkey` doesn't use it, but that's a minor issue for now.

Also, I'd refactor the `pkg-config` backed block to be a function so that the basically equal code doesn't have to be in `CMakeLists.txt` twice. If you don't want to do that though, that's okay, I can do that later, too. But it's always better to avoid copy-pasta, following the DRY principle, as that reduces the maintaining overhead and prevents spaghetti code.

TheAssassin:
Looks good. I'll test it later.

TheAssassin:
As previously mentioned, I assumed genkey doesn't require linking to zlib. A quick `ldd` call told me that I was right. CMake calls strip on the binaries it builds (except for when `CMAKE_BUILD_TYPE=Debug`), and it stripped that out.

So, actually `genkey_deps` doesn't need to be set. Could you please change that? Since my tests in Docker containers worked the way they should, I would merge it then.

Robalni:
The problem here is that the genkey source files include cube.h, which includes zlib.h. So if I uninstall zlib (and remove the zlib dependencies in CMakeLists.txt to avoid errors) then I can run cmake without errors but I can't build genkey because I don't have zlib.h. So zlib really is a dependency for genkey right now. One way to solve that would be to include tools.h instead of cube.h but to make that possible we would have to change tools.h to include what it needs because right now it relies on cube.h including everything.

TheAssassin:
Okay, that makes sense. And since we're not linking against it but just include the headers, it should be fine.

TODO: Refactor headers so that zlib is only included where necessary.

Thanks for your patch!

PR #699, opened 2017-06-19
Show users what version they need when loading demos and maps

Robalni:
I tried to load a demo but it told me that I needed an older version. I tried an older version but I still needed an older version... This change makes it possible to get the correct version without having to try different versions when loading demos and maps.


TheAssassin:
Merging, as there haven't been any complaints (or any other form of reaction) and it looks fine to me.

Issue #700, opened 2017-06-21
Red Eclipse disconnects on alt tab on KDE Neon 5.10


acerspyro:
The problem could also be seen on my previous installation with Gnome 3.24 on Ubuntu. Alt-tabbing outside of the game more often than not leads the game to disconnect from its currently connected server. Cannot be reproduced on Windows 10.

---
**First seen in**: stable version

**Observation**: Game disconnects from the server when alt-tabbing.

The game stops sending and receiving data when you alt-tab out of it, but only disconnects you 30 seconds after you alt-tab back in, leading me to think that the entire program was suspended while it was alt-tabbed out for some reason.

**How to reproduce**:
1. Launch the game under an Ubuntu KDE or Gnome desktop
2. Connect to a server
3. Alt-tab out and back into the game
4. Get disconnected

I searched the issue list and checked that this is not a duplicate :white_check_mark: 
I tried to rename my *config.cfg*, and the problem also occurs with default settings :white_check_mark:

pocak100:
I can't test these days so I won't send a patch, but what needs to be done is to get rid of `SDL_MinimizeWindow` in [engine/console.cpp:677](https://github.com/red-eclipse/base/blob/master/src/engine/console.cpp#L677) either for all platforms or for Linux only. That should take care of the disconnects.

Then, if Linux users complain about the game _not_ minimizing, take out [engine/main.cpp:1065](https://github.com/red-eclipse/base/blob/master/src/engine/main.cpp#L1064-L1066), so that RE respects the `SDL_VIDEO_MINIMIZE_ON_FOCUS_LOSS` environment variable, like other SDL games.

See also

acerspyro:
@pocak100 Alright, thanks. I will apply that patch and do my own testing.

PR #702, opened 2017-06-22
Fix alt-tab disconnecting the client from servers on some Linux systems

acerspyro:
See issue #700, fix provided by @pocak100

Tested on KDE Neon and Windows 10, works as intended on both.


TheAssassin:
Looks good now. @pocak100 can you please review this, too, please?

PR #705, opened 2017-07-05
Make oldschool duel/survivor count frags for points instead of rounds won.

shacknetisp:
I've seen this suggested several times before, it gives oldschool + duel/survivor a bit of a twist where you must try to frag as many players as possible before dying rather than only trying to survive.

### DIFF
--- a/src/game/duelmut.h
+++ b/src/game/duelmut.h
@@ -383,10 +383,10 @@ struct duelservmode : servmode
                                         {
                                             if(!m_affinity(gamemode))
                                             {
-                                                givepoints(clients[i], 1, true, teampoints);
+                                                givepoints(clients[i], 1, !m_dm_oldschool(gamemode, mutators), teampoints);
                                                 teampoints = false;
                                             }
-                                            else if(!duelaffin && teampoints)
+                                            else if(!duelaffin && teampoints && !m_dm_oldschool(gamemode, mutators))
                                             {
                                                 score &ts = teamscore(clients[i]->team);
                                                 ts.total++;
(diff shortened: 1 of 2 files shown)


JeffBobbo:
I've suggested this before and it didn't gain a lot of traction or interest, but it'd be nice that it actually worked or was marked as mutually exclusive, rather than not working as it is now.
https://redeclipse.net/forum/viewtopic.php?f=4&t=1079

shacknetisp:
Hmm, missed that forum post. Well, I'll leave it open for now, perhaps there is a way to improve the presentation.

JeffBobbo:
I think it's a good change, it means we can play survivor without rewarding hiding out to the end of the round.

Team survivor may pose an issue though, as you could stop the other team gaining points by committing suicide, e.g., committing suicide when on low health to stop the other team from gaining points, but you still get points for any kills you made.

Perhaps suicide should lose your team a point? In old school -1 for suicide is the same as the point the other team lost out on anyway.

shacknetisp:
With this PR suicide already does that, it behaves as in usual oldschool dm.

TheAssassin:
I'd suggest to combine both approaches, like, 1 kill = 1 point, but being the last one standing is a challenge, too, and should be credited with a point, too.

acerspyro:
So, from what I understand is this is a fight for a high score before you die rather than a fight to frag a lot?

shacknetisp:
It is a fight to frag a lot before you die.

JeffBobbo:
@acerspyro No, because it's not "a fight to frag a lot" anyway. In normal survivor you get points for being the last person alive. That can be achieved by hiding for the round and killing the penultimate person (meaning you get 1 point for 1 kill, and everyone else has done the work).

The behaviour of old school would be that you get a point for each kill, so in a 5 person ffa-ol-su dm game, there's 4 points available per round. Setting oldschool on survivor should shift the focus from survival into getting a high k:d. It doesn't matter if you don't win the round, as long as you take out >50% of the players, you'll come out on top.

TheAssassin:
As said before, we could credit both kills and surviving. What do you think?

PR #706, opened 2017-07-20
Clean up database. Improve players, weapons and servers schema.

zaquest:
Several updates to the stats db.

 * Removes invalid `game_weapons` and `game_weapons` that weren't used by a player (`timewielded=0 AND timeloadout=0`).
 * Removes `game_players` that weren't playing (`timeactive=0 AND timealive=0`).
 * Removes corrupted `games` that has several players with the same `game_players.wid`.
 * Removes data for non-existing games.
 * Add constraints to `game_servers`, `game_players` and `game_weapons` to improve database integrity, as currently there's a lot of invalid data.
 * Add table indices. This won't significantly improve `statsdb-interface` performance as there're other issues to it.

### DIFF
--- a/sql/stats/create.sql
+++ b/sql/stats/create.sql
@@ -14,14 +14,17 @@ CREATE TABLE games (
 );
 
 CREATE TABLE game_servers (
-    game INTEGER,
-    handle TEXT,
-    flags TEXT,
-    desc TEXT,
-    version TEXT,
-    host TEXT,
-    port INTEGER
+    game INTEGER PRIMARY KEY,
+    handle TEXT NOT NULL,
+    flags TEXT NOT NULL,
+    desc TEXT NOT NULL,
+    version TEXT NOT NULL,
+    host TEXT NOT NULL,
+    port INTEGER NOT NULL,
+    CHECK (handle <> ''),
+    FOREIGN KEY (game) REFERENCES games(id)
 );
+CREATE INDEX game_servers_handle ON game_servers(handle);
 
 CREATE TABLE game_teams (
     game INTEGER,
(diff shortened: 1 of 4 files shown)


swooboo:
In my opinion, this is at least 5 commits too little. Such a big change should not be a 1 single commit, for easier visibility and debuggability.

JeffBobbo:
@swooboo You realise that the number of commits is completely immaterial? Large changes just go in their own branch, which can be squashed into a single commit before merging.

shacknetisp:
The changes to the database in `upgrade_5.sql` must also be reflected in `create.sql` as new databases do not run the upgrade sequences. Otherwise this looks ok, though I don't know where the corrupted games are coming from.

TheAssassin:
> @swooboo You realise that the number of commits is completely immaterial? Large changes just go in their own branch, which can be squashed into a single commit before merging.

That is no justification for bad style of committing. And using squash to combine the commits before merging the changes is bad style, too. @JeffBobbo I'm of the same opinion as @swooboo here.

qreeves:
I really don't think it matters how many commits are in a PR, I prefer to "Squash and Merge" them anyway. @zaquest clearly outlined the changes made in his description, which is better than what most people manage.

zaquest:
I think the change is granular enough to be in one commit. It's basically "remove & prevent trash in the db". The list merely enumerates types of trash. Splitting it into commit per list item would produce unnecessary noise in both git history and `STATSDB_VERSION`.

shacknetisp:
I created a fresh version 5 database, but when launching a test master server I receive the following error:

```
...
Loading master (*:28800)..
Exiting: statistics database error: cannot start a transaction within a transaction
Server shutdown in progress..
```

shacknetisp:
Tested with interface and upgrading, all seems to be working now.

swooboo:
My argument is for visibility of changes and easier debugging. Commits should be as atomic as possible, which will make debugging easier, so `git bisect` will pinpoint the bug better. Also, I've never used squash and merge, since regular merge (without fast-forward) adds a special commit that unites the merged commits anyway. I'm pretty sure that 'commit smallest changes' is somewhere in Git user's manual.

PR #707, opened 2017-07-22
Fix for OSX, finding more reliably OpenGL as well.

devnexen:


src/CMakeLists.txt
```diff
@@ -118,8 +118,10 @@ endif()
 
 if(${BUILD_CLIENT})
     # add the client executable and link it to enet
+	find_package(OpenGL REQUIRED)
```
TheAssassin: Indentation fail. Also kind of cluttered now.

Also, I'd prefer to keep it use pkg-config. What are the practical advantages of using `find_package()` here?

src/CMakeLists.txt
```diff
@@ -137,13 +139,12 @@ if(APPLE)
...
     endif()
     target_link_libraries(redeclipse_server${BIN_SUFFIX} ${COCOA_LIBRARY})
-else(APPLE)
-    # add the genkey executable
-    add_executable(genkey${BIN_SUFFIX} ${genkey_sources})
-    set_target_properties(genkey${BIN_SUFFIX} PROPERTIES
-        COMPILE_FLAGS "-DSTANDALONE"
-    )
 endif(APPLE)
+# add the genkey executable
```
TheAssassin: This change implies that genkey can be built on Mac OS X. I don't remember why I put it in this else block. Anyway, looks good to me.

src/CMakeLists.txt
```diff
@@ -100,7 +100,7 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 elseif(APPLE)
     # build OS X specific Objective-C code
     file(GLOB mac_client_sources
-        xcode/main.m xcode/macutils.mm
```
TheAssassin: I'm curious. Why is `main.m` not required here? I basically rewrote our Makefile back in the days in CMake code. If it isn't needed to build the application, can we remove it from the source tree in total?

We need more information on this.

@qreeves might provide useful information, too.

src/CMakeLists.txt
devnexen: It does not find it on Mac

src/CMakeLists.txt
devnexen: Because both implements int main(...) so I get duplication of symbols

devnexen: Once @qreeves tells which one is the good to keep I can delete but I sense SDLmain is but maybe I m wrong.

TheAssassin:
You should avoid to `git pull --rebase && git push` in a pull request. Now we got duplicate commits in this branch and our master. We could probably only fix that right now by using GitHub's "squash & merge" feature.

devnexen:
Good point sorry for that.

src/CMakeLists.txt
TheAssassin: If I only had a Mac on hand, I could tell myself.

Let me try with [osxcross](https://github.com/tpoechtrager/osxcross) later.

TheAssassin:
I suddenly remembered why I preferred `pkg-config` over `find_package`. There were heavy issues without `pkg-config` when trying to cross-compile for Windows/macOS using MXE/osxcross/wclang/mingw/etc. We should keep that in mind, as cross compiling is a very convenient way to get binaries for other platforms.

TheAssassin:
@devnexen I would love to merge this soon, if you could revert your change on `find_package` as it breaks MXE builds (I just tested). Otherwise, thanks a lot, I don't know if anyone ever tested that on a "real" Mac, I just build it occasionally with osxcross.

TheAssassin:
Thanks!

qreeves:
Is it OK to merge this and test it on my Mac yet?

TheAssassin:
Oh, I thought you merged this already, @qreeves, but you just updated the branch.

In theory, yes, this can be merged, I don't see a problem. But for pure testing, how about cloning the branch directly before merging it?

```
git clone https://github.com/devnexen/base.git -b osx_build_fix
```

TheAssassin:
@devnexen the client won't build with osxcross:

```

**Comment by devnexen** [comment]
@TheAssassin 
cannot test right now buy could you possibly use SDL_main.* instead of main.m just to see if it works better ?

**Comment by devnexen** [comment]
I am surprised by your issue though I m not sure I introduced so many changes with this small PR.

**Comment by TheAssassin** [comment]
The issues in #720 have not been introduced by this PR. They are already in the master branch, into which PR #698 has been merged. I first had to fix the problems I had in #720 before I got the result I posted above. I then fixed them first before coming back to your suggestion.

I have not modified your branch at all. As far as I can see, @qreeves did not touch the code where you removed `main.m`, i.e. `SDLmain` is used indeed: https://github.com/red-eclipse/base/pull/707/files#diff-95e351a3805a1dafa85bf20b81d086e6L103

Thus, I come to the conclusion that the change breaks builds with osxcross, considering that after adding `main.m` again it builds just fine.
```
$ file redeclipse_osx 
redeclipse_osx: Mach-O 64-bit x86_64 executable

Issue #715, opened 2017-08-06
maxmerge is not a world variable


molexted:
The variable maxmerge, which is currently listed as simply "integer", controls the maximum cube size that the Cube 2 mipper can optimize the geometry to. It should be a world variable, saved to each map's config file, since specific maxmerge values can help alleviate performance problems and lighting errors.

Seen in 1.5.8 stable.

PR #716, opened 2017-08-06
Make maxmerge and minface world variables

acerspyro:
See https://github.com/red-eclipse/base/issues/715
Also my text editor seems to have removed a bunch of trailing spaces automatically, I hope this won't cause any problems.

### DIFF
--- a/src/engine/octa.cpp
+++ b/src/engine/octa.cpp
@@ -31,7 +31,7 @@ void setcubeext(cube &c, cubeext *ext)
     c.ext = ext;
     if(old) delete[] (uchar *)old;
 }
-  
+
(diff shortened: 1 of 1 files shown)


qreeves:
@lsalzman can you please confirm if this is OK to do, or if there are any caveats that make this not a good idea to implement?

lsalzman:
I guess the original intent was that this should be set in map cfgs explicitly. But technically it will still work as a world variable.

acerspyro:
Are there any other variables that can be set as a world variable?

TheAssassin:
Is that just me, or does @lsalzman's comment read like we shouldn't have merged it because the change doesn't provide what was expected?

PR #717, opened 2017-08-07
Give each weapon and hurt mat separate residual variables.

shacknetisp:
With this change each weapon & flak along with the hurt material can specify their own values for residuals (`hurtbleeddamage`, `zapperburntime2`, `shotgunflakshockstunscale1`, etc.)

This will allow for finer tuning of weapons and hurtmat, as an example it could allow the zapper to have less shock stun while the mine retains its current strength or a map author to have a very brief period of shock as a trap.

Naturally, testing will be appreciated.

This was requested specifically by AnDroidEl, and I have wanted this myself in the past.

_This pull request breaks protocol._


qreeves:
Please resync so I can merge.

PR #719, opened 2017-08-14
Use parentheses in macro

robalni:
I put parentheses in the macro mattrig so that things happen the expected way.

mattrig is called with a shorthand if statement in one place which messes up the `if (mw >= 0)` statement. gcc 7.1 also gives a warning about integer constants in boolean context.


TheAssassin:
Thanks for the patch, looks good to me, waiting for Semaphore before merging. I guess we might wish to add some builds for newer distros as well. Current builds are run on Trusty IIRC.

Issue #720, opened 2017-08-15
CMake `configure_deps` does not work as intended


TheAssassin:
CC @robalni 

As I noticed today while trying to compile with osxcross, the refactoring in #698 introduced an issue that is that `link_directories` is called too late.

`make VERBOSE=1` showed that the `-L<dir>` referring to the dependencies installed by `osxcross-macports` was missing. Albeit correctly returned by `x86_64-apple-darwin13-pkg-config --libs`, it wasn't included in the linker command. When adding it by hand, linking works fine (well actually it does not, see https://github.com/red-eclipse/base/pull/707#issuecomment-322429452).

I've developed a simple fix for that. I will create a PR as a platform for discussing this, as I'd like to hear @robalni's opinion on that.

PR #721, opened 2017-08-15
Fix CMake issues introduced in #698

TheAssassin:
As described in #720, some issues were introduced while refactoring the CMake configuration in #698.

I would further like to include osxcross toolchain I wrote for testing this, as it could be used for future cross compiling purposes.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,5 +1,30 @@
 cmake_minimum_required(VERSION 2.8.11)
 
+# link_deps([FATAL] DEPS <deps>...)
+# Calls link_directories for <deps>. <deps> is a list of dependencies.
+# If FATAL is specified then it exits on failure. Otherwise it returns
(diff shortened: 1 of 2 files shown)


robalni:
It looks all right to me. But I think the real problem is that we have to use link_directories at all, because pkg_check_modules doesn't return full paths. Is there any good reason we use pkg_check_modules instead of find_package?

TheAssassin:
`pkg-config` works with *any* pkg-config binary you hand to it. Check out the osxcross toolchain, osxcross comes with its own pkg-config that auto resolves paths to stuff installed conveniently with `osxcross-macports`. The same method works for MXE when building for Windows.

I explained this in #707 already.

If you can recreate this functionality without too much hassle, why not. But `pkg-config` seems to be the least hackish way.

TheAssassin:
Merging for now. We can replace `pkg-config` at any time in the future.

PR #725, opened 2017-09-01
Show message if deps not found in link_deps_dirs

robalni:
When `link_deps_dirs` was added it was used with the FATAL flag so that cmake would die without showing a message about using `-DBUILD_CLIENT=0` if the dependencies were not found. This happens because `link_deps_dirs` is executed before `configure_deps`, where the `configure_deps_failed` check is. In this PR this check is now performed in both places (but it is technically not needed in the second one anymore, should we keep it or remove it?).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -138,7 +138,10 @@ endif()
 
 if(${BUILD_CLIENT})
     # set up library directories
-    link_deps_dirs(FATAL DEPS ${client_deps})
+    link_deps_dirs(DEPS ${client_deps})
+    if(link_deps_dirs_failed)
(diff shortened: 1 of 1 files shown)


TheAssassin:
I have to admit I haven't invested much time into testing this feature yet. Even for Red Eclipse's AppImages I just download and build the entire game and then create AppImages from that for the client and the server, depending on whether either of them is enabled. But yes, I can reproduce the issue here.

I initially liked this feature because of being useful for [Red Eclipse's Docker containers](https://github.com/TheAssassin/redeclipse-docker). I haven't gotten to using it there as well, but will change that soon.

I'd say, go for whatever you prefer. I can live with both having the second check or removing it.

Thanks for your regular contributions by the way! They're greatly appreciated!

P.S.: I'll open a CMake related linking(major) issue soon, I'd appreciate your help there as well!

robalni:
Ok, I guess I will just leave it the way it is.

PR #726, opened 2017-09-02
Support both wget and curl in the *nix updater script.

### LINKED ISSUE #724: Updater scripts use brace completion in module prefetch url, incompatible with all but curl.
This currently only affects *nix without curl installed, although the root issue is present everywhere (including the windows batch file). When the updater tries to download the files from the variable `REDECLIPSE_MODULE_PREFETCH`, it will fail due to wget not understanding the brace syntax used to match multiple files in one command.

This affects both stable and master.

shacknetisp:
Here's a PR if you want to continue wget support for a while. Since there are (rare) circumstances where wget is installed but curl is not, this will increase potential compatibility while we continue using the current updating system.

`REDECLIPSE_DOWNLOADER` becomes a function:

* `REDECLIPSE_DOWNLOADER <output> <url>`

and the bulk-download form which downloads directly to `REDECLIPSE_TEMP`:

* `REDECLIPSE_DOWNLOADER "" <url> <url> <url>`

Fixes #724

### DIFF
--- a/bin/update.sh
+++ b/bin/update.sh
@@ -100,9 +100,9 @@ redeclipse_update_branch() {
             REDECLIPSE_DOWNLOADER()
             {
                 if [ -n "$1" ]; then
-                    wget --no-check-certificate -U "redeclipse-${REDECLIPSE_UPDATE}" -O "$1" "$2"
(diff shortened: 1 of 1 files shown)


acerspyro:
P.S. Ubuntu MATE 17.10 did not come with curl pre-installed.

TheAssassin:
@acerspyro this is what the PR attempts to mitigate. If curl is not available, it falls back to wget.

PR #730, opened 2017-09-20
Render non-lerp particles first. Fixes #728.

### LINKED ISSUE #728: Particles flickering when they are on a fog-heavy area.
Sometimes, let's say, you make a particle have a long time to fade out, and you got a very foggy area (fog 500/1000).
If you shoot, the particle just starts to flicker, it gets super bright, for no aparent reason.
It's related to fog, because it only occurs on that situation, but when a map has a big fog value (3000 ++ ), the particle doesn't start to glitch out.

The map from those screenshots for testing the issue ...

[particle-fog-bug.zip](https://github.com/red-eclipse/base/files/1309155/particle-fog-bug.zip)

This was already reported on this topic last year, but doesn't seem to be fixed, as on the last git version, it still occurs : https://redeclipse.net/forum/viewtopic.php?f=7&t=1054

I find this to be important to fix, as in theory, we could have some sort of "fake volumetric light" effect using a certain particle entity combination on certain maps (Stone in particular would benefit a lot, also could be used for atmospheric smoke particles on forests, like on my example map) , but with this bug, it isn't possible to do . (newent particles 15 100 100 0x999999 10000 5000 1 7 0 0 0 1000)

zaquest:


graphitemaster:
Why not keep two arrays of particles (one lerp, one non lerp) generate them in that order, avoiding the need to sort at all, and even if you didn't want to go that far, keep two sorted lists of each and avoid running through the particles twice, this seems like an inefficent fix otherwise.

graphitemaster:
yeah quite literally, just have two parts arrays ...

zaquest:
Sort is executed once during a game run, so this shouldn't be a performance issue. Having 2 arrays means 2 loops in renderparticles.

PR #731, opened 2017-09-29
Collide less with occluded faces. Fixes #221.

### LINKED ISSUE #221: wall run/clip bug
Seems there's some issues when transitioning from clip material to geometry where you get stuck kicking off an invisible wall.

http://dl.dropbox.com/u/14141289/RE/parkourbug.mp4

zaquest:
Most of the #221 comes from engine checking for collision surfaces that are occluded by it's neighbours. While in theory it doesn't seem feasable to eliminate all such cases, the combination of excluding fully occluded surfaces from collision detection and increasing rounding error margin fixes all the cases I've been able to find or construct.

### DIFF
--- a/src/engine/octa.h
+++ b/src/engine/octa.h
@@ -336,3 +336,4 @@ enum
     GENFACEVERTSXY(x0,x1, y0,y1, z0,z1, c0,c1, r0,r1, d0,d1) \
     GENFACEVERTSZ(x0,x1, y0,y1, z0,z1, c0,c1, r0,r1, d0,d1)
 
+bool collidesolidface(const cube &c, const int orient, const ivec &co, const int size);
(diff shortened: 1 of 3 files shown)


src/engine/octa.h
```diff
@@ -339,3 +339,4 @@ enum
     GENFACEVERTSXY(x0,x1, y0,y1, z0,z1, c0,c1, r0,r1, d0,d1) \
     GENFACEVERTSZ(x0,x1, y0,y1, z0,z1, c0,c1, r0,r1, d0,d1)
 
+bool collidesolidface(const cube &c, const int orient, const ivec &co, const int size);
```
qreeves: Should probably be `extern`

zaquest: Is it a code style thing? Otherwise there's no difference AFAIK.

TheAssassin: It's a linker thing, `extern` explicitly makes this symbol public. Even if it doesn't make an explicit difference linking the Red Eclipse binaries, this function should be made public in the library.

pocak100: Top level functions have [external linkage by default](http://en.cppreference.com/w/cpp/language/storage_duration#Linkage), so it makes no difference. It seems to be a convention in Cube 2 to mark them explicitly `extern` in header files.

acerspyro:
Looks fine to me. @qreeves @lsalzman please review.

Issue #722, opened 2017-08-19
Avoid spawning multiple players in the same spot


acerspyro:
This is a problem in games like Duel where two players would spawn on the same playerstart.

Additionally, we could use Quake's tactic and have players spawn at the playerstart that is the furthest away from as many people as possible on top of avoiding the same playerstart.

The first is pretty much a low-hanging fruit, but the second would be nice as well.

This is both a bug and a feature request. The first point definitively needs to be fixed.

shirepirate:
there are options in the entities for mutator-specific ones. but I recall finding out that this was broken when i tried to fix the duel maps near the release of 1.5. i don't know if they are working correctly now

qreeves:
A bug was identified in the spawn mechanics which was fixed. As a result, this should no longer be an issue.

PR #736, opened 2017-10-25
Break spawn protection on cooking and affinity taking actions.

shacknetisp:
Players under spawn protection can gain an advantage immediately after spawning by cooking weapons or taking an affinity while still invincible, but neither of these actions break their protection.
This pull request adds two on-by-default variables (`protectbreakcook`, `protectbreakaffinity`) that cause these actions to break protection, keeping spawn protection only as a method to assess the situation and reorient rather than affect the state of the match.

The old `protectbreak` variable is moved to `protectbreakshoot` for uniformity.

Requested by @JeffBobbo


TheAssassin:
I'm not so fond of switching this on by default. I don't think cooking a grenade should break the spawn protection, as this protects from spawn camping. If it broke spawn protection, one could just stand next to the spawn points, and wait for players to spawn and pick up a grenade. They can't effectively use it, as either way (cooking or throwing away) will break their protection.
I wouldn't switch that on by default, at least until the next release is published, similarly to the flag toss fix.

acerspyro:
See #722 
Implementing a Quake 3-like method of spawning would avoid most of these spawn killing issues anyway.

shacknetisp:
Players have three seconds after spawning to reposition themselves, they can grab explosives during this time without leaving their invincibility. Furthermore, if a player has recently died, they should not have an advantage over the attackers simply because they died.
I don't believe spawn camping is helped or hindered by this, the camper already has access to any nearby explosives, and the spawning player has those seconds of invincibility rendering any immediate attack useless. Players should not be using their spawn protection for offensive actions.

qreeves:
Please resync so I can merge.

PR #742, opened 2017-10-28
Revert "Include SDL and supporting libraries on linux due to broken SDL 2.0.6."

TheAssassin:
Reverts red-eclipse/base#741


TheAssassin:
We had to trigger a build to test this feature. The new libraries weren't bundled by the semaphore build, which lead to a broken binary. Reverting, waiting for feedback from @qreeves first.

PR #749, opened 2017-11-07
animated background

zaquest:
* Adds animated background shader `hudbackground`. The shader is needed early, so there's a copy in `setupshaders`.
 * Adds textures used by the background as variables `backgroundcausttex`, `backgroundwatertex` and `backgroundcloudtex`.
 * Removes `backgroundtex` variable.

### DIFF
--- a/config/glsl/hud.cfg
+++ b/config/glsl/hud.cfg
@@ -175,48 +175,46 @@ shader 0 "hudbackground" [
         pos = vec2(vvertex.x, vvertex.y);
     }
 ] [
-  const vec3 foreground = vec3(1.0, 0.0, 0.0);
-  const vec3 background = vec3(0.07, 0.0, 0.0);
-  const vec2 center = vec2(0.0, 1.5);
-  const float maxdist = distance(center, vec2(-1, -1));
-  const float waterweight = 1.2;
-  const float caustweight = 0.9;
-  const float cloudweight = 0.5;
-  const float timescale = 1.0 / 200.0;
+    const vec3 foreground = vec3(1.0, 0.0, 0.0);
+    const vec2 center = vec2(0.0, 1.5);
+    const float maxdist = distance(center, vec2(-1, -1));
+    const float waterweight = 1.2;
+    const float caustweight = 0.9;
+    const float cloudweight = 0.5;
+    const float timescale = 1.0 / 200.0;
 
-  const float totalweight = caustweight + waterweight + cloudweight;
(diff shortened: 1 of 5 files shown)


config/glsl/hud.cfg
```diff
@@ -165,3 +165,58 @@ shader 0 "hudcubemap" [
...
+
+      // bg
+      vec2 cuv = (pos + 1.0) / 2.0;
+      float cloud = cloudw * texture2D(tex2, cuv).r;
+      vec3 bg = foreground * cloud;
+
+      float k = log(dist) / log(maxdist);
+      fragcolor = vec4(mix(fg, bg, k), 1);
+  }
+]
```
qreeves: This block has incorrect indentation.

src/game/hud.cpp
```diff
@@ -1651,17 +1655,40 @@ namespace hud
...
+            settexture(backgroundcloudtex, 0x300);
+            glActiveTexture_(GL_TEXTURE0);
+
+            t = NULL;
             mapbg = 0;
         }
-        glBindTexture(GL_TEXTURE_2D, t->id);
-        float offsetx = 0, offsety = 0;
-        if(showloadingaspect&(1<<mapbg))
+        if(t)
```
qreeves: Redundant, may as well be `else` with the following `else` block put above.

src/game/hud.cpp
```diff
@@ -1651,17 +1655,40 @@ namespace hud
...
         if(!t || t == notexture)
         {
-            t = textureload(backgroundtex, 3);
+            (hudbackgroundshader ? hudbackgroundshader : nullshader)->set();
+
+            pushhudmatrix();
+            hudmatrix.ortho(-1, 1, -1, 1, -1, 1);
+            flushhudmatrix();
+
+            LOCALPARAMF(time, lastmillis / 1000.0f);
```
qreeves: Depending on what you want here, using `lastmillis` will not animate when the game is loading something as the game won't update this value until it gets back to the main loop. You *can* use `SDL_GetTicks()` for something that does update, but then it might be choppy as the screen is updated irregularly during loading sequences.

Not an issue, just an FYI so you can make your own choice.

zaquest: Yes, I don't think it's feasable to provide a non-stop animation here. My objective was just to prevent it from flickering and time skipping. So that it stayed still when something else is happening and it continued from the place where it stopped.

PR #751, opened 2017-11-08
Remove hardcoded shaders from `engine/shader.cpp:setupshaders()`.

zaquest:
Add a `updateprogress` flag to `setupshaders`. Move "early" shaders to a separate file. Remove their copies from the source code.


qreeves:
Looks like @lsalzman took care of this after seeing this PR 😛

PR #767, opened 2017-11-19
Let mouse leave game's window. Close #701.

### LINKED ISSUE #701: Missing button to release the mouse
Older versions of the game used to feature a mouse release button on the F9 (?) key.

The fact that this key is now missing is extremely irritating in conjunction with the fact that the game automatically minimizes itself upon alt-tab. This prevents me from right-clicking its titlebar to change the window to Fullscreen in KDE. Even if I can get the menu to open via Alt + F3, it cannot be controlled, as the game also captures the keyboard.

The best fix would be to let the mouse leave the window if it goes beyond the edge of the window.

zaquest:
This works by not grabbing mouse for anything but camera control and syncronizing system's and game's cursors when necessary.

### DIFF
--- a/src/engine/main.cpp
+++ b/src/engine/main.cpp
@@ -47,13 +47,12 @@ void textinput(bool on, int mask)
 
 VARN(IDF_PERSIST, relativemouse, userelativemouse, 0, 1, 1);
 
-bool shouldgrab = false, grabinput = false, canrelativemouse = true, relativemouse = false;
+bool grabinput = false, canrelativemouse = true, relativemouse = false;
 
 void inputgrab(bool on)
 {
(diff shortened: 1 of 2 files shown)


acerspyro:
Could you make the in-game cursor disappear when the mouse leaves the window?

lsalzman:
This looks like it will break grabbing and also not properly interact with alt-tabbing and such on all operating systems. Not a fan of this.

qreeves:
Yeah, we need to be really careful with the input stuff due to individual operating system quirks. It is on my list of things to look at though.

Calinou:
Also, make sure it still works with multi-monitor setups.

PR #770, opened 2017-11-21
New team balance style: average position on the scoreboard

shacknetisp:
This new balance style (`teambalancestyle 7`) is the average of the player's score compared to the maximum score (i.e., their usual position on the scoreboard), so that a player always at the top will have a score of `1.0`, always in the middle will be `0.5`, and always at the bottom will be `0.0`. These averages are taken from the last 50 games (`masterstatsavgposlastgames 50`) the player has played in with `uniqueplayers >= 4` (`masterstatavgposnumplayers 4`) and `normalweapons`.

Server-local statistics count too, at a 33% weight (`teambalanceavgposlocalweight 0.33`) for authed players and 100% weight for unauthed players since they don't have global statistics.
Local statistics are updated every 5 seconds (`teambalanceavgposdelay 5000`) as they are counted as an average and required the highest score to be calculated.

Example scores for some well-known players, you can see the overall accuracy here:

| Handle | AvgPos |
| ------ | ------ |
| acerspyro | 0.490 |
| bobbo | 0.763 |
| dziq | 0.373 |
| molexted | 0.373 |
| qreeves | 0.311 |
| snisp | 0.665 |
| zaquest | 0.595 |

This style is superior to kdratio as it counts scoreboard position, which includes objectives rather than only frags, and is therefore applicable to every game mode. It is set as the default in this pull request.

_This pull request breaks protocol._


qreeves:
Will need a `CLCOMMAND` in `game/client.cpp` in order to access `clientstate::totalavgpos`.

You can also increment `GAMEVERSION` for this PR.

PR #839, opened 2018-10-19
Use pkg-config instead of freetype-config and sdl-config

reinerh:
freetype-config is deprecated and in some distributions even no longer
shipped:
https://git.savannah.gnu.org/cgit/freetype/freetype2.git/commit/?id=b0a9383
The recommended way to find the library and compilation flags is
using pkg-config.

As sdl2-config has similar issues, use pkg-config for SDL2 as well.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -10,6 +10,8 @@ PLATFORM= $(shell $(CC) -dumpmachine)
 
 RMFLAGS= -fv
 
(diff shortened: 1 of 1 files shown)


qreeves:
Is `pkg-config` present in most distributions now?

reinerh:
I would expect so, I'm not aware of any where it's not present.

PR #840, opened 2018-10-23
fix cmake steam linking for linux

IceflowRE:
This fixes at least the cmake building for linux.
Iam not sure if there are better fixes, because this cmakelist is a mine field.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -116,7 +116,13 @@ set_source_files_properties(support/sqlite3.c PROPERTIES
 # platform specific code
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     list(APPEND client_deps x11)
-    link_libraries(rt)
+    add_library(steam_lib SHARED IMPORTED)
+    if(${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "i[3-6]86")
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -116,7 +116,13 @@ set_source_files_properties(support/sqlite3.c PROPERTIES
...
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     list(APPEND client_deps x11)
-    link_libraries(rt)
+    add_library(steam_lib SHARED IMPORTED)
+    if(${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "i[3-6]86")
+        set_target_properties(steam_lib PROPERTIES IMPORTED_LOCATION ${CMAKE_CURRENT_SOURCE_DIR}/../bin/x86/libsteam_api.so)
+    elseif(${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "x86_64")
+        set_target_properties(steam_lib PROPERTIES IMPORTED_LOCATION ${CMAKE_CURRENT_SOURCE_DIR}/../bin/amd64/libsteam_api.so)
+    endif()
+    link_libraries(rt steam_lib)
```
qreeves: Is this correct? The name is `libsteam_api` and usually referenced as `steam_api`

IceflowRE: `steam_lib` is just the variable name i used to reference.

PR #849, opened 2018-12-29
add support for obj model vertex coloring

q009:
This change adds support for vertex coloring in .obj models.
While non-standard, this method is adopted by a few modeling applications and will allow for an extra layer of compatibility.

The functionality may (and should) be extended by doing the same for .iqm models, since it has vertex color attributes as well.

While currently not used, this change allows us to pass additional per-vertex information to shaders.

This change is required for the next PR I'll be putting up which introduces procedurally animated vegetation via vertex shaders.

(Manually backported from my Tesseract fork. While I haven't noticed any, please be on the lookout for potential mistakes).

### DIFF
--- a/src/engine/animmodel.h
+++ b/src/engine/animmodel.h
@@ -694,6 +694,20 @@ struct animmodel : model
                 lastbbuf = lastvbuf;
             }
         }
+
+        void bindcolor(void *v, int stride)
+        {
+            if(!enablecolor)
+            {
+                gle::enablecolor();
+                enablecolor = true;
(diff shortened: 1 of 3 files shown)


lsalzman:
This is not done in an efficient way. Every model pays for color whether it uses it or not. It should not be bloating the vertex format when it is not used.

q009:
I had this in mind, I guess I shrugged it off for later.
Very well, I'll take care of this.

q009:
Changes applied.

PR #850, opened 2018-12-30
Procedural vegetation animation

q009:
This change introduces procedural vegetation animation.
With it, it's no longer necessary to manually animate trees, etc.

In order to use this functionality your model needs to have the following information supplied in vertex colors:

* red channel - leaf wiggliness
* green channel - animation phase shift
* blue - detail (branch) sway

In addition, you must enable it via 'mdlvegetation 1' in your model config.

There's a neat piece of software (suggested by Ardelico) TreeIt. It can quickly generate very high quality vegetation and can export to .obj, which then can be loaded up in the game in order to see the new functionality working out-of-the-box (with a provided config).

https://www.youtube.com/watch?v=8CEIbPGJ_ro

### DIFF
--- a/config/glsl/model.cfg
+++ b/config/glsl/model.cfg
@@ -62,9 +62,63 @@ skelanim = [
     ]
 ]
 
+// Cubic Bezier periodic wave function
+cbezier = [
+    result [
+        float cbezier(float x) { return (mod(x, 2) * (mod(x, 2) - 1) * (mod(x, 2) - 2)) * 2.5; }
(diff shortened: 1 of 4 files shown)


q009:
Thanks. Changes applied.

q009:
Seems to me the shader is the only piece of code that would actually benefit from comments so I've added a few. Also, got rid of magic numbers. While it goes against the convention it makes it significantly easier to fine tune the shader.

PR #856, opened 2019-01-20
Unlock FOV values for firstpersonfov and thirdpersonfov

acerspyro:
This PR is to improve the consistency of the four FOV variables.

Currently, the Edit mode and Spectator mode FOV values are unlocked and go from 1 to 179 degrees, while the firstpersonfov and thirdpersonfov values are locked from 90 to 150 degrees.

The variables are now locked at the UI level to avoid problems with users rendering their game unplayable.


Calinou:
I believe the lower FOV cap for `firstpersonfov`/`thirdpersonfov` was added to decrease the efficiency of zoom scripts. (That said, due to how FOV is calculated in Cube 2, the wider your aspect ratio, the more efficient a zoom script that uses FOV 90 will be.)

acerspyro:
@Calinou Thank you for your input. Do you think all FOV variables should be locked to be between 90 and 150 degrees instead? Anything further than that is rather useless in most if not all cases.

Besides, someone could simply re-compile their game with an unlocked variable and use the cheat script anyway.

Calinou:
> @Calinou Thank you for your input. Do you think all FOV variables should be locked to be between 90 and 150 degrees instead? Anything further than that is rather useless in most if not all cases.

The current FOV limits are fine to me, but perhaps we should automatically calculate the FOV for a 4:3 aspect ratio and increase it for wider aspect ratios (or switch to vertical FOV, which gives the same result). This way, the default FOV would be correct on widescreen displays like in all modern games ("[Hor+](http://www.wsgf.org/article/screen-change)").

acerspyro:
> > @Calinou Thank you for your input. Do you think all FOV variables should be locked to be between 90 and 150 degrees instead? Anything further than that is rather useless in most if not all cases.
> 
> The current FOV limits are fine to me, but perhaps we should automatically calculate the FOV for a 4:3 aspect ratio and increase it for wider aspect ratios (or switch to vertical FOV, which gives the same result). This way, the default FOV would be correct on widescreen displays like in all modern games ("[Hor+](http://www.wsgf.org/article/screen-change)").

I completely agree. A separate PR / Issue should be opened about this.

I will be updating this branch shortly to keep values for all four variables between 90 and 150 degrees.

shacknetisp:
Is this ready to be merged?

acerspyro:
This is ready to be merged as far as I'm concerned.

shacknetisp:
Please edit the title/description to reflect what's actually going being changed here, it looks like the purpose has changed entirely from the original.

PR #861, opened 2019-01-29
Adjust textures UI to reflect vrotate changes

acerspyro:
Added support for `/vrotate 6` and `/vrotate 7` in the textures UI


molexted:
Probably OK to merge at this point unless you have second thoughts about the nomenclature after three weeks.

PR #864, opened 2019-02-16
new vegetation

q009:
Currently for new vegetation.


shacknetisp:
Does anything need added to `config/`?

q009:
No, it's entirely self-contained.
If a user wants to use the assets they will need to execute the supplied configs themselves.

molexted:
Users won't likely execute the repository because they likely won't realize it's been added. Recommend adding entries to config/map/models.cfg.

q009:
Very well

shacknetisp:
Merged.

PR #873, opened 2019-02-28
Reimplement editing quit and disconnect warnings in the menu.

### LINKED ISSUE #871: Reimplement quit confirmation.
The quit confirmation (e1ea5654d38385f4e580dbe656b91620a3e2eae1) has been removed due to #811, but should be reimplemented properly.

Relevant commits removing it: 703d909, c8d108d

shacknetisp:
Closes #871 

Original implementation: e1ea5654d38385f4e580dbe656b91620a3e2eae1

The difference is adding the actual disconnect/quit commands if the ui is displayed outside of edit mode and a minor removal of magic mode number 1 in favor of `modeidxediting`.

### DIFF
--- a/config/ui/main.cfg
+++ b/config/ui/main.cfg
@@ -38,8 +38,8 @@ uimenu "main" "Main Menu" "textures/icon" [
                 uibuttonm "Server Browser" "textures/servers/list" [uiopen "servers"] 0 main_servers
                 uibuttonm "Settings" "textures/icons/settings" [uiopen "settings"] 0 main_settings
                 uibuttonm "Help and Support" $questiontex [uiopen "help"] 0 main_help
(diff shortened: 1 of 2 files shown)


molexted:
I see that the quit popup appears when trying to quit from the main menu (and that this appears to be intended behavior), but the wording is a bit off, since the quit message notes that `All unsaved changes will be lost.` It may be worth having a quit UI and a editquit UI so that it doesn't spook people into thinking they're going to lose their settings or something.

Minor nitpick though, and everything looks and works OK otherwise.

shacknetisp:
> I see that the quit popup appears when trying to quit from the main menu (and that this appears to be intended behavior)

That wasn't intended.

molexted:
Seems to work good now with the recent fixes.

PR #886, opened 2019-05-06
Change included makefile to use pkg-config for all dependencies.

lambda-11235:
This should allow RE to compile on systems that have different package locations then the Makefile expects. I'm not sure why only SDL2 was in a package config to begin with.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -122,8 +122,8 @@ else
 BIN_SUFFIX=_native
 endif
 endif
-CLIENT_INCLUDES= $(INCLUDES) -I/usr/X11R6/include `$(PKG_CONFIG) --cflags sdl2`
-CLIENT_LIBS= -Lenet -lenet -L/usr/X11R6/lib -lX11 `$(PKG_CONFIG) --libs sdl2` -lSDL2_image -lSDL2_mixer -lz -lGL
+CLIENT_INCLUDES= $(INCLUDES) -I/usr/X11R6/include `$(PKG_CONFIG) --cflags x11 sdl2 SDL2_image SDL2_mixer zlib gl`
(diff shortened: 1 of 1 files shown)


acerspyro:
Which system(s) can we test this on so we can approve and pull the PR?

IceflowRE:
Did not semaphore already checked it? Afaik semaphore does not check cmake only.

pocak100:
Is `-L/usr/X11R6/lib` still necessary?

qreeves:
I think there was a reason we didn't use `pkg-config` for all of the dependencies, but I don't remember why anymore. As I don't use Linux, I can't properly review this PR.

> Is `-L/usr/X11R6/lib` still necessary?

I think it used to be used for things like clipboard, but SDL2 has access to that now. Upon review, it looks like we're still using Xlib for capslock/numlock status.

pocak100:
I meant that `pkg-config` will provide a path for Xlib (`libX11.so`) if it's in a special location. (On my system, it's in `/usr/lib`, and there's no `/usr/X11R6` directory.) Thus, no need to hardcode the path unless maybe something other than Xlib needs it. Same for the include path.

>I think there was a reason we didn't use `pkg-config` for all of the dependencies, but I don't remember why anymore. As I don't use Linux, I can't properly review this PR.

Merge and see what breaks? :)

qreeves:
Does `pkg-config` provide `-L` (library directory) and `-I` (include directory) as well? If so, they should be removed as well before I merge.

pocak100:
Yes, as I understand it, that's the point of pkg-config.

qreeves:
You'll need to revert the changes to `enet` as we use a bundled version, and the author of it is @lsalzman (a developer for Cube 2, Tesseract, and Red Eclipse), who discourages using the versions in Linux distributions.

PR #896, opened 2019-08-03
vector-based wind and emission entities

q009:
This change introduces vector-based wind entities and emitters.
The shader will now animate meshes in the direction of the wind, which is either provided by temporary emitters or map entities.

Introduced map variables:
windspeed(alt)
windyaw(alt)
windinterval(alt)

New map entity "wind"

Please consult the in-game info/docs for usage.


src/engine/wind.cpp
```diff
@@ -0,0 +1,255 @@
...
+    // apply the direction
+    vecfromyaw(checkmapvariant(MPV_ALT) ? windyawalt : windyaw, 1, 0, globalwind);
+    globalwind.mul(speed);
+}
+
+// gets the wind for dynamic (moving) entities
+static vec getentwindvec(const dynent *d)
+{
+    // add dynent velocity to the wind vector
+    return d ? vec(d->vel).add(d->falling).mul(-WIND_ENT_FALL_SCALE) : vec(0);
```
q009: Noticed this oopsie here, this is wrong

qreeves: That moment when people review their own pull requests 🤣

qreeves:
You think you're able to add `wind.[cpp|o]` to the other project files (CBP, XCODEPROJ), or do you need me to do it?

src/engine/wind.cpp
```diff
@@ -0,0 +1,255 @@
+#include "engine.h"
+#include "game.h"
+
+#define WIND_IMPULSE_DECAY_MUL 4.0f
+#define WIND_ENT_FALL_SCALE 0.01f
+#define WIND_ATTEN_SCALE 0.01f
+#define WIND_MAX_SPEED 2.0f
+#define WIND_MAX_EMITTERS 100
```
qreeves: Is this stuff that could be a variable?

q009: I don't think exposing those to the user would provide any value.

q009:
> You think you're able to add `wind.[cpp|h|o]` to the other project files (CBP, XCODEPROJ), or do you need me to do it?

Projects updated

PR #907, opened 2019-08-31
revert to dynamic allocation in N_UNDO handling

### LINKED ISSUE #841: undo causing segfaults
When undoing a fairly large geometry change of a map (seems to happen more often at "gridpower 10" and above), the game will either freeze or segfault, with no other information relating to error messages or logs.

eg; `Fatal signal 11 (Segmentation Violation)`

Here's a reproducible method;

* /edit cathedral
* /gridpower 10
* select a specific area of the map (see image below)
* press backspace to delete selection
* press Z to undo
* game then crashes with segfault

2018-11-01-181646_1920x1080_scrot

This map is not specific to the issue, just one in the default map rotation that can reproduce it. Although, the problem seems to be related to the number of world triangles in the selection, the more geometry that is present, the more likely the game will crash.

` Version: 1.9.9-nix64-inplace client (v2 Preview) [0x3c603210] (Dist)`
Although, i recall having this happen a few times in 1.6 also, but could never reproduce it, not sure if this has been reported before, although SniperGoth just told me they've experienced it a few times.

Hopefully someone else can reproduce this form the above methods?

pocak100:
Fixes #841.

The size of undo packets is unbounded in single player.

### DIFF
--- a/src/game/server.cpp
+++ b/src/game/server.cpp
@@ -7569,8 +7569,7 @@ namespace server
                         break;
                     }
                     if(p.remaining() < packlen) { disconnect_client(sender, DISC_MSGERR); return; }
-                    uchar s[MAXTRANS];
-                    ucharbuf q(s, MAXTRANS);
+                    packetbuf q(32 + packlen, ENET_PACKET_FLAG_RELIABLE);
                     putint(q, type);
                     putint(q, ci->clientnum);
                     putint(q, unpacklen);


qreeves:
Aha, I see. Naughty @zaquest 😄 
```
$ git blame -L 7572,7573 src/game/server.cpp
ca207abf5 (zaquest 2016-12-01 20:19:36 +0700 7572)  uchar s[MAXTRANS];
ca207abf5 (zaquest 2016-12-01 20:19:36 +0700 7573)  ucharbuf q(s, MAXTRANS);
```

Dziq84:
Earlier game crashes with undoing one specific test-prefab I have,  now its all OK :)
...but game still crashes when undoing many ents ;p

PR #910, opened 2019-09-07
Tweak `maxfps` to be friendly to 144 Hz displays

Calinou:
These new default values result in a smoother user experience for owners of 144 Hz displays, as they won't have to anything specific to use their display to its full extent.

### DIFF
--- a/src/engine/main.cpp
+++ b/src/engine/main.cpp
@@ -645,7 +645,7 @@ void swapbuffers(bool overlay)
 }
 
 VAR(IDF_PERSIST, menufps, 0, 60, 1000);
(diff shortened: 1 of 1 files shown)


molexted:
Does setting `menufps` to 144 even help with anything? I worry that it would default laptops to burning more power on the menus than it needs to.

Calinou:
@molexted It makes the mouse cursor movement smoother on 144 Hz displays, but I agree it's kind of wasteful otherwise. Ideally, we'd have a way to check the monitor's refresh rate and set the FPS cap to match it.

PR #911, opened 2019-09-07
Adjust the `maxfps` menu slider

Calinou:
The minimum value has been raised, as very low `maxfps` values make menus difficult to control.

The maximum value has been raised to account for 240 Hz displays, which have the highest frequency available on the market as of this commit.

### DIFF
--- a/config/ui/settings.cfg
+++ b/config/ui/settings.cfg
@@ -339,7 +339,7 @@ uimenu "settings" "Settings" "textures/icons/settings" [
                                         uialign -1 0
                                     ]
                                     uihlist $ui_padbutton [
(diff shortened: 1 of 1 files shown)


qreeves:
The problem with this is that `0` is valid as meaning "no limit" and recent changes make `-1` use the refresh rate (which is also now the default setting). Care to review?

Calinou:
@qreeves I'd like to expose `0` and `-1` as values in the slider, but this means the slider would have to allow very low FPS values to be used (including `1`, which makes the GUI unusably slow). Maybe we should have checkboxes below the slider for "Unlimited" and "Use monitor refresh"?

qreeves:
Yeah, that's up to you 😄

PR #926, opened 2019-10-02
Add pingservers command and use it in server UI init

### LINKED ISSUE #915: Server list takes a while before displaying server information
- Version and build: 1.9.9 4906fbbb1c2c88bbac6271df4949a9a161aa641b
- Observation: The server list won't display server information until a certain delay has passed.
- Expected result: The server list should request server information immediately, then request is periodically while the list is open.
- How to reproduce: Open the server browser. With the default settings, it will take up to 10 seconds for server information to appear after the list has been received. By "server information", I mean the player counts, current map, …

This bug has been present in Red Eclipse since forever, but I thought it'd make sense to report it now :slightly_smiling_face:

Now that the server refresh delay no longer seems to be exposed to the GUI, it's no longer possible to work around it easily (without using the console, that is).

molexted:
Fixes #915 

Changes:
- adds `pingservers` CS command which justs executes the `pingservers()` that the server update loop already uses
- automatically executes this and `updateservers` on server UI open, 1s apart, so the list appears populated immediately even at long update intervals (which may still be desirable)

### DIFF
--- a/src/engine/animmodel.h
+++ b/src/engine/animmodel.h
@@ -525,7 +525,7 @@ struct animmodel : model
         template<class V, class TC, class T> void calctangents(V *verts, TC *tcverts, int numverts, T *tris, int numtris, bool areaweight)
         {
             vec *tangent = new vec[2*numverts], *bitangent = tangent+numverts;
-            memset(tangent, 0, 2*numverts*sizeof(vec));
+            memclear(tangent, 2*numverts);
             loopi(numtris)
             {
                 const T &t = tris[i];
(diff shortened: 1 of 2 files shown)


src/engine/serverbrowser.cpp
```diff
@@ -297,6 +297,7 @@ void pingservers()
     }
     lastinfo = totalmillis;
 }
+COMMAND(0, pingservers, "");
```
qreeves: I don't recommend exposing `pingservers` in this fashion. I think the better option is for `refreshservers()` to take a bool argument that overrides `lastinfo` time check which gets provided during `updatefrommaster()`, so this only happens directly after the list is first retrieved.

PR #927, opened 2019-10-08
Make prefab and mapmodel UIs more user friendly

lambda-11235:
Updated prefab and mapmodel UIs (F5 and F6 in edit mode) to have a search and preview.


Dziq84:
I even don't know that prefab UI exists in 1.9.   Molex why it's not in editing menu ;p?

qreeves:
Gonna close this because it will become part of the wider edit mode redesigning we're doing now.

PR #928, opened 2019-10-08
[Wip] Update Cmake

IceflowRE:
This PR will merge a completely reworked CMake.
Which finally allows to build on Windows, Linux and MacOS.
With these changes the discord-rpc library will be updated to 3.4.0 and discord and steam .lib for windows are added.

## Notes
- `src/include/` directory is not used
- `src/xcode/main.m` is not used
- `bin/redeclipse.app/Contents/Frameworks` SDL2 framework folders may not used
- `src/lib/` the .lib for sdl2, zlib and enet may not be required

### Linux
```
# base tools
pacman -S cmake make gcc pkgconf
# dependencies
pacman -S zlib freetype2 sdl2 sdl2_image sdl2_mixer
# create cmake build
cmake -B./cmake-build-debug/ -DCMAKE_BUILD_TYPE=Debug ./
# build
cmake --build ./cmake-build-debug/
```

### Windows
You need a Mingw installation like MSYS2.
```
# base tools
pacman -S mingw-w64-x86_64-gcc mingw-w64-x86_64-cmake mingw-w64-x86_64-pkg-config
# dependencies
pacman -S mingw-w64-x86_64-zlib mingw-w64-x86_64-SDL2 mingw-w64-x86_64-SDL2_mixer mingw-w64-x86_64-SDL2_image mingw-w64-x86_64-freeglut mingw-w64-x86_64-glew mingw-w64-x86_64-freetype
# create cmake build
cmake -B./cmake-build-debug/ -DCMAKE_BUILD_TYPE=Debug -G "MinGW Makefiles" ./
# build
cmake --build ./cmake-build-debug/
```

### MacOS
```
# install brew
/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
# base tools
brew install git cmake pkg-config
# dependencies
brew cask install xquartz
brew install zlib sdl2 sdl2_image sdl2_mixer freeglut freetype
# create cmake build
cmake -B./cmake-build-debug/ -DCMAKE_BUILD_TYPE=Debug ./
# build
cmake --build ./cmake-build-debug/
```

### Control what to build
By default every part will be build, you can turn of specific builds with cmake options. Same about Steam and Discord support.

```
-DBUILD_CLIENT=OFF
-DWANT_STEAM=OFF
-DWANT_DISCORD=OFF
-DBUILD_SERVER=OFF
-DBUILD_CUBE2FONT=OFF
-DBUILD_GENKEY=OFF
```

## TODO:
- [ ] Test Linux -> Windows cross compiling


src/CMakeLists.txt
```diff
@@ -1,284 +1,338 @@
...
+        DESTINATION ${CMAKE_CURRENT_SOURCE_DIR}/../bin/${RE_ARCH}/
     )
-    add_executable(cube2font${BIN_SUFFIX} ${cube2font_sources})
-    set_target_properties(cube2font${BIN_SUFFIX} PROPERTIES
-        COMPILE_FLAGS "-DSTANDALONE"
+endif ()
+
+if (${BUILD_GENKEY})
+    add_executable(
+        redeclipse_genkey
```
qreeves: Just `genkey.exe` and `genkey_linux` with no prefix please.

src/CMakeLists.txt
```diff
@@ -1,284 +1,338 @@
...
+    set_target_properties(redeclipse_genkey PROPERTIES OUTPUT_NAME "genkey")
+    install(
+        TARGETS redeclipse_genkey
+        DESTINATION ${CMAKE_CURRENT_SOURCE_DIR}/../bin/${RE_ARCH}/
+    )
+endif ()
+
+if (${BUILD_CUBE2FONT})
+    add_executable(
+        redeclipse_cube2font
```
qreeves: Same as with `genkey`.

src/CMakeLists.txt
IceflowRE: The executable name here ist just the internal reference name. The **output** name is defined later, with `set_target_properties`

WatIsDeze:
Sweet

src/CMakeLists.txt
```diff
@@ -1,284 +1,338 @@
...
 
+# set C++ standard
+set(CMAKE_CXX_STANDARD 11)
 # set compile options
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -fsigned-char -ffast-math -fno-exceptions -fno-rtti -Wno-invalid-offsetof")
 
+# add Clang Thread Safety Analysis
+if (__clang__)
+    add_compile_options(-Wthread-safety)
+endif ()
```
pocak100: As far as I can tell, this needs annotations in the source code to do anything, which we don't have.

src/CMakeLists.txt
```diff
@@ -1,284 +1,338 @@
...
+option(BUILD_CLIENT "Build client?" ON)
+option(WANT_STEAM "Build client with Steam support?" OFF)
+option(WANT_DISCORD "Build client with Discord support?" OFF)
+option(BUILD_CUBE2FONT "Build cube2font?" ON)
+option(BUILD_SERVER "Build server?" ON)
+option(BUILD_GENKEY "Build genkey?" ON)
+
+# disable steam and discord support if it is not possible to link them
+if (NOT ${BUILD_CLIENT})
+    message(WARNING "Disabling Steam and Discord support as client is not build")
```
pocak100: *built

qreeves:
> One nitpick: it doesn't add the `_linux` suffix to executables. I don't mind, but maybe Quin does.

It's mostly because when BSD binaries used to be built, they got the `_bsd` suffix. In the end, it is best to replicate what the Makefile and deploy servers are doing anyway.

src/CMakeLists.txt
IceflowRE: Upps, will be fixed.

src/CMakeLists.txt
IceflowRE: Will look up.

IceflowRE:
Will change the suffix.

src/CMakeLists.txt
IceflowRE: Removed as the required "notes" are not used anyway yet.

IceflowRE:
@redeclipse/developers 
Ready for review.

src/CMakeLists.txt
```diff
@@ -1,284 +1,367 @@
...
+        support/jsmn.c
+        support/sqlite3.c
 
-# neither server nor client need cube2font.cpp - to avoid warnings about duplicate main()s, it has to be removed from their source lists
-file(GLOB cube2font_cpp_path shared/cube2font.cpp)
-list(REMOVE_ITEM client_sources ${cube2font_cpp_path})
-list(REMOVE_ITEM server_sources ${cube2font_cpp_path})
+        redeclipse.rc
+    )
+    set_source_files_properties(redeclipse.ico redeclipse.rc PROPERTIES LANGUAGE RC)
```
pocak100: This line seems to make having a .rc compiler a build requirement. As in, cmake gives an error if you don't have one. AFAICT, the .rc and .ico files are only used on Windows. Could you please make it so that this property is only set when compiling for Windows?

qreeves:
The Discord and Steam shared libraries need to be removed from this PR as well, they have nothing to do with the CMake update.

src/CMakeLists.txt
IceflowRE: Yeah was thinking about this too, but actually it didnt harm on other systems as cmake takes care of.
So still a windows only or can it be like now?

pocak100: I did some reading and I'm still perplexed. Why is this line (225th) here?

If the rc language is *enabled*, cmake should recognize the .rc file extension, and have it compiled and linked into the executable.
If it's *not* enabled, then the extension is not recognized, and the file ignored, so no problem on Linux/Mac.
But if you set the file's language manually, you're telling cmake the file *must* be compiled, [according to the documentation](https://cmake.org/cmake/help/latest/prop_sf/LANGUAGE.html), and that will of course fail if the user doesn't have the relevant compiler. We shouldn't require Linux users to install `windres`.

As for the .ico, doesn't it get picked up by the resource compiler? I notice that it's not in the sources list.

So, could you check again if the line is needed at all, and if it is, wrap it in an `if`?

Or, have a look at [`check_language`](https://cmake.org/cmake/help/latest/module/CheckLanguage.html), if the problem is rc not always being enabled when compiling for Windows.

src/CMakeLists.txt
```diff
@@ -1,284 +1,367 @@
...
+set(RE_ARCH_64 OFF)
+set(RE_ARCH_ppc64 OFF)
+set(RE_ARCH_ppc OFF)
+set(RE_ARCH_arm OFF)
+set(RE_ARCH_aarch64 OFF)
+set(RE_ARCH_NATIVE OFF)
+if (${CMAKE_SYSTEM_PROCESSOR} MATCHES "amd64.*|x86_64.*|AMD64.*")
+    set(RE_ARCH "amd64")
+    set(RE_ARCH_64 ON)
+elseif (${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "i686.*|i386.*|x86.*")
```
pocak100: I just noticed that this `if` is not checking the same variable as the previous. I think the one without `HOST` is the right one.

IceflowRE: oooopppss silly copy paste mistake 🤦‍♂

IceflowRE: `CMAKE_SYSTEM_PROCESSOR` is the right value.
> The name of the CPU CMake is building for.

IceflowRE:
@qreeves Removed the discord-rpc update and just added the .lib for windows building.
@pocak100 fixed the mentioned mistakes.

IceflowRE:
Applied tessfont from upstream.

PR #943, opened 2019-11-16
Fix blood stains

pocak100:
The code base is a bit confused about whether stain type should use a zero-based or one-based index. That leads to an off-by-one error that makes falling blood particles leave scorch marks on the floor. This patch rectifies that by standardizing on a one-based index with zero meaning no stain.

As a side note, `STAIN_SMOKE` is not actually used anywhere in the code base.


src/shared/iengine.h
```diff
@@ -334,12 +334,13 @@ extern void regularflame(int type, const vec &p, float radius, float height, int
 // stain
 enum
 {
-    STAIN_SMOKE = 0,
+    STAIN_NONE = 0,
```
q009: I'd probably continue to use a 0-based indexing approach, STAIN_NONE being defined as -1, because now the type known as STAIN_NONE still has to have a valid stainrenderer defined, which is a bit counter-intuitive I'd say.

pocak100: There's no `stainrenderer` corresponding to `STAIN_NONE`. When looking up the `stains` array, you have to offset the type by one, as c++ arrays are zero based. Said lookup only happens in `::addstain()`, so it's not burdensome.

Standardizing on zero indexing would be difficult. For example, if you create a new particle entity, it will have all attributes set to zero, so that would mean having some decal by default, and that's undesirable. Existing maps would have to be fixed up somehow too.

And having both types of indexing is clearly error prone.

qreeves: I looked into this and @pocak100 is correct.

src/engine/stain.cpp
```diff
@@ -809,8 +809,8 @@ VAR(IDF_PERSIST, maxstaindistance, 1, 512, 10000);
 
 void addstain(int type, const vec &center, const vec &surface, float radius, const bvec &color, int info)
 {
-    if(!showstains || type<0 || (size_t)type>=sizeof(stains)/sizeof(stains[0]) || center.dist(camera1->o) - radius > maxstaindistance) return;
-    stainrenderer &d = stains[type];
+    if(!showstains || type<=0 || (size_t)type>sizeof(stains)/sizeof(stains[0]) || center.dist(camera1->o) - radius > maxstaindistance) return;
+    stainrenderer &d = stains[type-1];
```
qreeves: I think here, it'd be better to start the function with:
```
int id = type-1;
```
then use:
```
stainrenderer &d = stains[id];
```
and leave the `if` statement alone, so it is clearer what is going on.

src/shared/iengine.h
```diff
@@ -334,12 +334,13 @@ extern void regularflame(int type, const vec &p, float radius, float height, int
...
     STAIN_SCORCH, STAIN_SCORCH_SHORT,
     STAIN_BLOOD,
     STAIN_BULLET,
     STAIN_ENERGY,
     STAIN_STAIN,
+    STAIN_SMOKE,
```
qreeves: Is there a reason you moved `STAIN_SMOKE`?

pocak100: It couldn't stay at index 0, obviously, and shifting down all the enum values by one would have put their values out of sync with particle emitters in existing maps.

src/engine/stain.cpp
pocak100: Yep, looks better.

PR #944, opened 2019-11-17
Smooth hudgun sway

q009:


src/game/game.cpp
```diff
@@ -3312,8 +3328,8 @@ namespace game
         mdl.flags = flags;
         mdl.basetime = mdl.basetime2 = 0;
         mdl.size = size;
-        mdl.yaw = d->yaw;
-        mdl.pitch = d->pitch;
+        mdl.yaw = (third == 1) ? d->yaw : swayyaw;
+        mdl.pitch = (third == 1) ? d->pitch : swaypitch;
```
qreeves: I'm assuming you want to add sway to the legs, in which case you should use just `third ? a : b` as `third = 2` is this pass.

q009: Thing is, the firstperson passes are on third 0 and third 2.
third 1 is for thirdperson

PR #946, opened 2019-12-05
bomber ball fix: successfully passed ball not getting claimed by recipient

pocak100:
### Problem
Sometimes, as the ball is passed among players, it hovers around the recipient without getting taken by them. Even though it's right on top of them.

### Cause
To see if a player can pick up the ball, the game checks the distance between it and the player's foot. However, a passed ball homes towards the player's head, which is far enough from the foot (especially with the recent player size increase) to not trigger a pickup.

### Proposed solution
Check if the ball is within the cylinder that encompasses the whole player.

### DIFF
--- a/src/game/bomber.cpp
+++ b/src/game/bomber.cpp
@@ -579,10 +579,16 @@ namespace bomber
         if(d->ai) aihomerun(d, d->ai->state.last());
     }
 
+    static inline bool pointinsidecylinder(vec pt, const vec center, float height, float radius)
+    {
+        if(pt.z < center.z-height/2 || pt.z > center.z+height/2) return false;
+        pt.sub(center);
+        return pt.x*pt.x + pt.y*pt.y <= radius*radius;
+    }
+
     void checkaffinity(gameent *d, int i)
     {
         if(!(AA(d->actortype, abilities)&(1<<A_A_AFFINITY))) return;
-        vec o = d->feetpos();
         bomberstate::flag &f = st.flags[i];
         if(f.owner)
         {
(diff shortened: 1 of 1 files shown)


src/game/bomber.cpp
```diff
@@ -608,7 +614,9 @@ namespace bomber
         }
         if(!f.droptime && m_bb_assault(game::gamemode, game::mutators) && d->team == T_ALPHA && bomberassaultreset) return;
         if(f.lastowner == d && f.droptime && lastmillis-f.droptime <= bomberpickupdelay) return;
-        if(o.dist(f.pos()) <= enttype[AFFINITY].radius/2) client::addmsg(N_TAKEAFFIN, "ri2", d->clientnum, i);
+        float radius = enttype[AFFINITY].radius;
+        if(pointinsidecylinder(f.pos(), d->o, radius/2, d->feetpos().z-radius/4, d->feetpos().z+d->height+radius/4))
```
IceflowRE: Could you not only pass the height to the function and calculate bottom and top inside?

pocak100: `bool pointinsidecylinder(vec pt, const vec bottomcenter, float height, float radius)`

`if(pointinsidecylinder(f.pos(), d->feetpos().sub(vec(0,0,radius/4)), d->height+radius/2, radius/2))
`

Not sure if it looks better.

qreeves: Maybe it'd be better to use a ellipsoid and remain more consistent with the way it previous worked? It's also probably perfectly fine to calculate from `d->center()` as well rather than these workarounds for using the feet position. I might have to look into this further for the other intersection checks (CTF, DAC, entities, etc).

pocak100: I've looked at using `ellipsecollide()` directly, but it modifies some global variables so I dared not. I figured the difference from a cylinder wouldn't be noticeable, so I went with that, since it's simple.

Then I went looking for what defines the top and bottom of the player entity (in a way that respects the resize mutator etc.), and somehow arrived at `feetpos()` and `feetpos() + height`. I guess I should have just asked. So, `center().z +- height/2` is it?

Maybe we should consider the radius of the player as well.

If you want to do a better fix with ellipses, I don't mind if you close this. I just thought it's better to not leave things broken for the v2 release, even if bb is not an official mode.

qreeves: Yeah, that's right. Go ahead and update the PR if you want, I may not have time to circle back to reviewing something different for it before the release.

PR #947, opened 2019-12-05
bomber ball: show target's name above reticle

pocak100:
It's hard to see the name tags on distant players, especially at the default `aboveheadnamessize`. Because of that, I often find myself accidentally passing the bomber ball to a bot instead of a human player.

This patch puts the name of the target above the reticle so you can know to whom you're about to pass the ball.

It fades out when you're close, disappearing roughly when the normal name tag is the same size.

### DIFF
--- a/src/game/bomber.cpp
+++ b/src/game/bomber.cpp
@@ -204,6 +204,8 @@ namespace bomber
                     gle::colorf(rp, gp, bp, sq*0.5f);
                     hud::drawsized(sx+ss/4, sy+ss/4, ss/2);
                 }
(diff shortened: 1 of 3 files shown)


qreeves:
Might be better to use something like `camera1->o.distrange(e->o, game::affinityhintfadeat, game::affinityhintfadecut)` as this would use the pre-existing systems and account for proper distance, not just depth in the scene. Maybe make your own variables for this scenario.

pocak100:
OK, I wasn't aware of those variables, will look into that.

pocak100:
Thanks for the hint, `distrange()` is perfect.

The values of `affinityhintfade*` weren't ideal, so I made new variables.

The thresholds now scale with `aboveheadnamessize`.

Issue #52, opened 2015-02-15
Player animation improvement suggestions


Calinou:
* don't play animation when player is not actually moving (speed = 0),
* play animation slower when player walks/runs slower than default walking/running speed, and faster when they walk/run faster than the default speed.

qreeves:
Second part of this needs @lsalzman to adjust this model rendering to support arbitrary speeds set by the software.

acerspyro:
We also need someone who actually understands how the Red Eclipse player model works, I never managed to export it correctly.

PR #969, opened 2019-12-22
Only display moving animations when the player is actually moving

Calinou:
This partially addresses #52.


Calinou:
Is this good to merge now, or is something blocking this?

qreeves:
I have my concerns on how this will look and if it should only be applied to some of these cases.

That being said, I'll merge it so everyone can evaluate and discuss further.

PR #971, opened 2019-12-23
Disable shadows for most dynamic lights

Calinou:
Most dynamic lights are small and only last for a short time, so it makes sense to disable shadows for better performance.

Moreover, this prevents some dynamic light effects from intersecting with the first-person/third-person player models, which can look bad.

Feel free to review the list of dynamic lights I tweaked. Maybe we could leave them enabled for DAC bases, but I'm not entirely sure. The fact you can see DAC base lights through thin walls can be beneficial to gameplay, even if it's technically incorrect.

### DIFF
--- a/src/game/bomber.cpp
+++ b/src/game/bomber.cpp
@@ -396,7 +396,7 @@ namespace bomber
             int millis = lastmillis-f.displaytime;
             if(millis <= 1000) trans = float(millis)/1000.f;
             vec colour = isbomberaffinity(f) ? pulsecolour() : vec::fromcolor(TEAM(f.team, colour));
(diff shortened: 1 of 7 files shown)


q009:
This shouldn't be outright disabled. Firstly, I'd be ideal to have some sort of distance check.
Secondly, there should be an option to force the shadows off for dynamic entities and also completely off.

Jigoku:
If anything illumradius/illumlevel probably shouldn't cast shadows, as the light source is buried inside the playermodel mesh.

qreeves:
`illum*` is supposed to be like a torch, and with the new engine we can actually do it properly now.

PR #974, opened 2019-12-24
substring -> substr

acerspyro:
``substr`` is noted as ``substring`` in the latest master, causing UI stuff to break and spam the console.


acerspyro:
Confused red-eclipse with redeclipse, don't mind me.

PR #976, opened 2019-12-25
fix `musictype 1` selecting the wrong mapmusic

### LINKED ISSUE #967: "musicedit 1" always plays random track
Problem:
`musicedit 1` ignores the track set within `mapmusic`. Despite the description of:

> mapmusic (or random if none)

Reproduce:
* set a mapmusic
* set `musicedit 1`
* random track is played

Jigoku:
Fixes #967

- add undocumented musicedit option to usage.cfg
- play correct mapmusic for `musictype 1`


src/game/game.cpp
```diff
@@ -3022,7 +3022,7 @@ namespace game
                 else
                 {
                     defformatstring(musicfile, "%s", mapmusic);
-                    if(musicdir[0] && (type == 2 || type == 5 || ((type == 1 || type == 4) && (!musicfile[0] || !fileexists(findfile(musicfile, "r"), "r")))))
+                    if(musicdir[0] && (type == 2 || type == 5 || ((type == 1 || type == 4) && (!musicfile[0] || fileexists(findfile(musicfile, "r"), "r")))))
```
qreeves: I'm not sure how this fixes the issue, `fileexists` returns `true` if the file exists, and it should only resort to random if it doesn't?

Jigoku: Just reapplied the patch now, and checked again; each musictype option behaves more like expected:

| musictype        |  (master)           | (this pull request)  |
| - | - | - |
| 0 = no in-game music | :heavy_check_mark:  | :heavy_check_mark: 
| 1 = map music (or random if none) | always random | plays mapmusic if set (random if none)
| 2 = always random | :heavy_check_mark:  | :heavy_check_mark: 
| 3 = map music (silence if none) | :heavy_check_mark:  | :heavy_check_mark: 
| 4-5 = same as 1-2 but pick new tracks when done | always random | plays mapmusic *

* 4 and 5 plays the mapmusic, then has a brief pause, before looping the same track (when it should be switching to random after mapmusic finishes).

Maybe something else needs changing there too as for 4 and 5, but with `!filexists`, all mapmusic is ignored as for the default `musictype 1` setting.

PR #979, opened 2019-12-26
Don't emit player boost particles while paused

Calinou:
This prevents the game from becoming very slow after pausing for a long time.

For the record, this isn't done using `canemitparticles()` as the boost effect seems to intentionally bypass the check (presumably to avoid having its emission frequency limited by the `emitmillis` variable).

### DIFF
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -1056,6 +1056,10 @@ namespace game
 
     void boosteffect(gameent *d, const vec &pos, int num, int len, bool shape = false)
     {
+        // Prevent spawning particles while the game is paused, as it doesn't clear particles in this state.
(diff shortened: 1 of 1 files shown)


src/game/game.cpp
```diff
@@ -1056,6 +1056,12 @@ namespace game
 
     void boosteffect(gameent *d, const vec &pos, int num, int len, bool shape = false)
     {
+        // Prevent spawning particles while the game is paused, as it doesn't clear particles in this state.
+        // Otherwise, the number of particles could become very large, causing the renderer to slow down to a crawl.
+        if (paused) {
+            return;
+        }
```
qreeves: Should conform with coding standards:
- No space between function and brackets.
- No unnecessary braces `{}`.
```C++
if(paused) return;
```

PR #980, opened 2019-12-26
Improve crosshair health display

Calinou:
The new implementation has numerous upsides compared to the previous implementation:

- It indicates the player's health more accurately using a color gradient well-known among FPS players. Health above 100
  is indicated by tinting the crosshair green.
- It's always visible, even in screenshots. Due to lack of animation, it also avoids hiding the health status on every loop of the animation (since the crosshair became white).

The variable was renamed from `crosshairflash` to `crosshairhealth`
to reflect the change.


src/game/hud.cpp
```diff
@@ -1174,11 +1174,31 @@ namespace hud
...
+                    // Yellow-white
+                    c.r = 1.0f;
+                    c.g = 1.0f;
+                    c.b = (ratio-0.6f)/0.4f;
+                } else {
+                    // Green (overheal)
+                    c.r = 2.0f - ratio;
+                    c.g = 1.0f;
+                    c.b = 2.0f - ratio;
+                }
```
qreeves: Should conform with coding standards:
- No space between function and brackets: `if (...)`
- No inlined with braces (`} else if (...) {`)
```C++
if(ratio < 0.2f)
{
    ...
}
else if(ratio < 0.6f)
{
    ...
}
```

q009 (approved):
I think this is fine, given it's gated behind a variable. So, it's up to each user's decision whether they wish to use this mechanism or not.

qreeves (changes requested):
I've looked at this more closely, and my issue with it is that it replaces the flash with the colour completely. I, personally, don't like the crosshair changing colour, which is why this was implemented as a flash instead that only alerts to low health.

Also the variable being used was only renamed in the `usage.cfg` and not in the source code. I suggest it be a bitwise variable: `&1` = flash, `&2` = colour. This way you can combine the two effects if you want as well.

PR #981, opened 2019-12-26
Print the modified map CRC message only when the map CRC differs

Calinou:
This reduces noise in the console when many players join a server.


molexted:
Possibly also handle the case of no map (which IIRC is a zero'd out CRC) while you're at it.

qreeves:
This only really seems to substitute one message with another? You'll still get a flood of CRC messages, just without the server CRC appended. As far as I'm aware, these should be going to the debug console anyway, which is hidden and separated by default.

molexted:
These messages go to the event console, so they're just as visible as the match start/end/OT messages.

PR #982, opened 2019-12-26
Increase the client/server update rate to 125 Hz

Calinou:
**This needs playtesting before it can be merged.** I can host a 24/7 server with this pull request if this is desired :slightly_smiling_face:

This decreases effective latency significantly by making clients and servers sending and receiving updates sooner. This increases network traffic by a factor of 3, but it should be manageable on today's networks.

The average game-induced latency (on top of network lag) should now be 8 ms instead of 40 ms.

See https://extra-a.github.io/#bnet for more information.


qreeves:
Needs further testing and discussion.

q009:
A lot of games actually have this as a setting. I'd probably go the same route.

q009 (changes requested):
The idea appears worth exploring to me. However, I think it'd be worth to put those behind variables, so the users can tune those. So, worst case scenario, the gist of the change may remain, only the final value of the variable would see further tweaks.

PR #983, opened 2019-12-28
Improve ping coloring in the server browser

Calinou:
There are now 5 possible colors instead of just 3. This makes for a smoother gradient.

### DIFF
--- a/src/engine/rendertext.cpp
+++ b/src/engine/rendertext.cpp
@@ -284,6 +284,7 @@ COLOURVAR(pink, 0xFF8080);
 COLOURVAR(violet, 0xB060FF);
 COLOURVAR(purple, 0xFF00FF);
 COLOURVAR(brown, 0xA05030);
+COLOURVAR(chartreuse, 0xB0FF00);
 
 static void text_color(char c, bvec4 *stack, int size, int &sp, bvec4 &color, int r, int g, int b, int a, int flags)
 {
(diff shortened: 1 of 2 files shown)


src/engine/rendertext.cpp
```diff
@@ -304,6 +305,7 @@ static void text_color(char c, bvec4 *stack, int size, int &sp, bvec4 &color, in
         case 'v': case 'V': stack[sp] = color = TVECA(c == 'V' ? colourdarkviolet : colourviolet, alpha); break;
         case 'p': case 'P': stack[sp] = color = TVECA(c == 'P' ? colourdarkpurple : colourpurple, alpha); break;
         case 'n': case 'N': stack[sp] = color = TVECA(c == 'N' ? colourdarkbrown : colourbrown, alpha); break;
+        case 'h': case 'H': stack[sp] = color = TVECA(c == 'H' ? colourdarkgreenyellow : colourgreenyellow, alpha); break;
```
Calinou: I picked `h` as a letter randomly, maybe there's a better choice?

molexted: Could use C for *chartreuse* if that's not conflicting with anything, I suppose.

Calinou: `c` is already used for cyan above :slightly_frowning_face:

qreeves: `H` seems fine, I'd also change the name `COLOURVAR` to `chartreuse`.

PR #984, opened 2019-12-29
Return iphostlock

voidanix:
Something went terribly wrong, why expose such a valuable piece of information?

1.6 did this but it involved operators, and the GDPR is pretty clear about it:
```
The GDPR states that IP addresses should be considered personal data as it enters the scope of ‘online identifiers’.
```

So: hide IP addresses unless you are an admin or higher ranked


qreeves:
It has been determined the GDPR in this instance only applies if there isn't a _legitimate interest_ in said information: as it interferes with the proper administration and moderation of the game, it has been disabled by default and won't be changed back.

Due to the divisive nature of arguments over the GDPR, I will have to lock this thread, as the issues on `base` are not the appropriate forum for this kind of discussion.

PR #986, opened 2019-12-29
Don't display the `/0` ammo store on the HUD for grenades/rockets/mines

Calinou:
This also fixes the "infinite ammo store" condition to make the pistol HUD display correct again.

## Preview

Grenade

Pistol


IceflowRE:
Can we add a real infinite symbol too? or is that the real one? Looks more like a rotated 8.

molexted:
That's because it's a rotated `8`.

PR #989, opened 2020-01-02
Make edit mode entity ellipses more detailed

Calinou:
This makes edit mode ellipses look less "angular".

## Preview

### DIFF
--- a/src/engine/renderparticles.cpp
+++ b/src/engine/renderparticles.cpp
@@ -1047,12 +1047,16 @@ struct loopprimitiverenderer : listrenderer<loopprimitive>
 
     void renderpart(loopprimitive *p, int blend, int ts, float size)
     {
+        // The number of polygon sides to use.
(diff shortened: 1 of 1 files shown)


q009 (approved):

src/engine/renderparticles.cpp
```diff
@@ -1047,12 +1047,16 @@ struct loopprimitiverenderer : listrenderer<loopprimitive>
 
     void renderpart(loopprimitive *p, int blend, int ts, float size)
     {
+        // The number of polygon sides to use.
+        // The polygon rendering breaks if this is set too high.
+        const int detail = 40;
```
qreeves, line 1052: I think I would rather this be a variable. There are legitimate performance reasons to not want ellipses to be highly detailed as there can be large amounts of them if you select a lot of entities.

qreeves (changes requested):
Changes requested.

PR #990, opened 2020-01-03
generalize architecture handling a little

q66:
Now ppc and aarch64 systems are properly detected and used in both the build system and the launcher, there ismore reliable detection for the makefile via a shell script (using compiler triples, which match across systems) and so on.

It's not perfect, but better what we have right now.

Also, do not set -ffast-math by default on platforms other than x86 where it's confirmed safe. On other targets it may generate unsafe code that will not behave correctly.

### DIFF
--- a/redeclipse.sh
+++ b/redeclipse.sh
@@ -51,11 +51,20 @@ redeclipse_setup() {
             i486|i586|i686|x86)
                 REDECLIPSE_ARCH="x86"
                 ;;
+            ppc64*|powerpc64*)
+                REDECLIPSE_ARCH="ppc64"
+                ;;
+            ppc*|powerpc*)
+                REDECLIPSE_ARCH="ppc"
+                ;;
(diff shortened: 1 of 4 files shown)


qreeves:
Fails to pass build tests: https://semaphoreci.com/redeclipse/base/branches/master/builds/3022

q66:
I don't see how that could've possibly affected build in that way, considering it actually makes the condition for `-lsteam_api` *stricter* than before.

qreeves:
Yeah, I can't figure out what's wrong with it either, but Windows 64 bit needs to use `-lsteam_api64`.

q66:
Ah, I see what's wrong, `PLATFORM` is special-cased for cross-compiling scenarios. It shouldn't really be (it should be able to automatically figure this out from the compiler), but basically just merge this but without the bits changing `64` to `x86_64` for mingw stuff for the time being

PR #999, opened 2020-01-12
vangle/texangle commands for fine texture rotation

q009:
`vangle` and `texangle` allow rotation of textures in degrees.
A helper function exists that automatically calculates vangle based on the slope: `autotexangle`

### DIFF
--- a/config/glsl/shared.cfg
+++ b/config/glsl/shared.cfg
@@ -175,4 +175,11 @@ screentexcoord = [
     ]
 ]
 
+rottexcoord = [
+    result [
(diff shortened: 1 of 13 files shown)


q009:
I still wonder whether it makes sense to bump the game version so soon, even if we added a new vslot property.

qreeves:
Yeah, it'll have to wait for the next point release I think.

q009:
I neglected to mention that there's an alternative way to implement this. A way that doesn't require any source code changes.

Instead of introducing a new vparam, one could simply define a new shader param and pass sin/cos values to it directly via setshaderparam. The only problem with this approach is the fact that shader parameters are not synchronized in coop, which is why I chose not to take this route.

qreeves:
`vshaderparam`?

PR #1013, opened 2020-02-09
regen begin sound

q009:


qreeves:
I'll need you to resync this one for me, had multiple sounds ones in a row.

PR #1018, opened 2020-02-15
spline interpolation for rails

q009:


src/game/entities.cpp
```diff
@@ -92,6 +92,13 @@ namespace entities
             if(parents.find(n) < 0) parents.add(n);
         }
 
+        rail &getrail(int cur, int offset)
+        {
+            int index = cur + offset;
+            if(index < 0) index += rails.length();
+            return rails[index % rails.length()];
```
qreeves: Your wrap-around point needs to be `ret` after the first iteration. The purpose of this variable is to allow the track to loop back in on itself to a place other than first rail, allowing the initial section of track to play only on the first iteration.

```
o<-<-o<-<-o<-<-o
|         |
o->->o->->o
```

q009: Ah, you're right

PR #1021, opened 2020-02-23
check for null pointers returned by lookupshaderbyname

### LINKED ISSUE #1008: segfaults relating to lava material
Struggling to find the reason for this, but there are now several people experiencing this strange one.

Essentially what happens (confirmed in edit mode, not sure about others), is that when you jump into lava material, the game will crash with signal 11. It can be reproduced "most" times by placing a playerstart above a lava pit. When the client connects to the server, and the map is fetched, the game will segfault immediately. It happens very regularly, but not all of the time. However, i'm confident it's related to the player entering lava material.

I don't have any more info than that, because each time it's occurred (offline, and online with multiple clients)... i've gone and built a debug build to try and grab a backtrace, but the game refuses to crash as soon as you launch a debug build. So it's a bit of a mystery for now.

pocak100:
`lookupshaderbyname()` may return NULL if the shader failed to compile (GLSL error), or the shader just wasn't preloaded yet.

Also, it turns out the engine uses the `waterfog` shader when the camera is inside lava. However, `waterfog` is only loaded if the map has water. This PR preloads it for lava as well.

Fixes #1008.

### DIFF
--- a/src/engine/material.cpp
+++ b/src/engine/material.cpp
@@ -413,6 +413,7 @@ void setupmaterials(int start, int len)
     if(hasmat&(0xF<<MAT_LAVA))
     {
         useshaderbyname("lava");
+        useshaderbyname("waterfog");
         loopi(4) if(hasmat&(1<<(MAT_LAVA+i))) lookupmaterialslot(MAT_LAVA+i);
     }
     if(hasmat&(0xF<<MAT_GLASS))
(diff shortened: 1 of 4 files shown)


Jigoku:
This patch also appears to also fix https://github.com/redeclipse/base/issues/1009 by the looks of it?

pocak100:
Nope, it only works against segfaults. I can't reproduce #​1009.

PR #1030, opened 2020-03-12
break stuff

voidanix:
Another thing that could be done would be removing the stats code to remove sqlite3.c or whatever it's called because it's full of holezz


IceflowRE:
If you change the makefile you should do the same change to the CMake one.

PR #1034, opened 2020-03-31
allow point lights to be linked to triggers

Jigoku:
Allows light entities to be directly linked to a trigger. 

There are a few use cases for this feature, one being that light sources for a sword weapon spawn (eg;volumetric light above the weapon platform) can be hidden when $mutsbitclassic isn't set. 

I'm not sure if i've missed anything, but seems to work fine from my testing locally.

### DIFF
--- a/src/engine/light.cpp
+++ b/src/engine/light.cpp
@@ -40,7 +40,7 @@ GETSKYPIE(lightpitch, float);
 
 bool getlightfx(const extentity &e, int *radius, int *spotlight, vec *color, bool normalize, bool dyncheck)
 {
-    if(!entities::isallowed(e)) return false;
+    if(!entities::isallowed(e) || !e.spawned()) return false;
     if(dyncheck && e.flags&EF_DYNAMIC) return false;
     if(color)
     {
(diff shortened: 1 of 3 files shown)


Jigoku:
Found a problem with this PR, existing maps end up with no lights emitting anything (even a /newent light). However, only a newmap being saved/loaded will work. Not sure why that happens though.

q009:
I'd squash those commits into one

qreeves:
I will get to this, but I have to be a bit more thorough with things that mess with entities and triggers.

Jigoku:
fixed conflicts and using getlightfx() as suggesed for the spawned check

q009 (approved):
LGTM

qreeves:
I'd rather leave these mediated through the `LIGHTFX` entity for now and put this on the backburner until a proper trigger system rewrite.

Jigoku:
> I'd rather leave these mediated through the `LIGHTFX` entity for now and put this on the backburner until a proper trigger system rewrite.

it was fun playing with the concept, but i do agree with that.

PR #1044, opened 2020-04-29
add UI setting to control hud padding

Jigoku:
Was suggested on discord that this could be useful for ultrawide monitors;

- allows horizontal/vertical scaling of the HUD to be closer to the centre of the screen
- add sliders to User Interface settings in the menu


Calinou:
As a 32" monitor user, I think this is a great feature :slightly_smiling_face:

PR #1045, opened 2020-05-02
animmodel vertex color fixes

q009:
This change makes sure the vertex colors are always bound when a model needs them (e.g. has wind enabled).

Also, introduced 'debugvertcolors 0/1' to view vertex colors on wind animated models in-game.

### DIFF
--- a/config/stdlib.cfg
+++ b/config/stdlib.cfg
@@ -286,3 +286,5 @@ decalload = [
 autotexangle = [
     vangle (atan (divf $arg1 8))
 ]; setcomplete autotexangle 1
+
+defvar debugvertcolors 0 0 1 resetgl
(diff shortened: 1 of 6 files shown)


config/stdlib.cfg
```diff
@@ -286,3 +286,13 @@ decalload = [
 autotexangle = [
     vangle (atan (divf $arg1 8))
 ]; setcomplete autotexangle 1
+
+dbgvertcol = 0
```
qreeves, line 290: Maybe `defvarp dbgvertcol 0 0 1` would be more appropriate, as it behaves like a real variable rather than an alias.

q009: Neat, changes applied

PR #1052, opened 2020-05-06
Request dedicated GPU when starting Red Eclipse from the `.desktop` file

Calinou:
See <https://www.hadess.net/2020/05/dual-gpu-support-launch-on-discrete-gpu.html> for more information about this newly added `.desktop` entry property.


qreeves:
If code can be written for the WM to do this on Linux, then maybe it should be hard coded into the game itself instead of a `.desktop` (which isn't used by default or on Steam versions) like it already is for Windows?

Calinou:
@qreeves We can do this for open source graphics drivers by setting the `DRI_PRIME=1` environment variable early enough. Unfortunately, the NVIDIA proprietary driver ignores it.

See https://github.com/godotengine/godot/blob/058a0afdeca83145d58a95c426dd01216c397ea9/platform/linuxbsd/detect_prime_x11.cpp for an example. It seems SDL doesn't do this on its own (I searched its repository for instances of `DRI_PRIME`, it has zero).

PR #1061, opened 2020-05-25
Remove movie feature

voidanix:
Arguably unused


Dziq84:
is there a reason why this should be removed? oO

voidanix:
The feature is unused by most people, the defaults give a low quality recording and changing them might unexpectedly throw a `computer too slow` error, stopping the recording

A dedicated recorder is better at handling this

Dziq84:
xD

PR #1065, opened 2020-06-09
purge mapmodels

dc-de:
mapmodels for purge


qreeves:
OK, so we can't just add a folder to this directory. We either need to create you a submodule or you can create a directory in the `misc` submodule. Closing this and you can let me know which you'd prefer on Discord if you like.

PR #1069, opened 2020-06-19
add clearconsole & clearchat commands

Jigoku:

- add `/clearconsole <console id>` command
- add `/clearchat` command
- add command usage to usage.cfg
- renamed the previous `clear_console()` to `clear_binds()` as it was only doing `keyms.clear();`

### DIFF
--- a/src/engine/console.cpp
+++ b/src/engine/console.cpp
@@ -713,10 +713,18 @@ void processkey(int code, bool isdown)
     else if(!consolekey(code, isdown) && !hud::keypress(code, isdown) && haskey) execbind(*haskey, isdown);
 }
 
-void clear_console()
+void clear_binds()
 {
     keyms.clear();
 }
+
+void clear_console(int type)
+{
+   if (type >= 0) conlines[type].clear();
+   else loopi(CON_MAX-1) conlines[i].clear();
+}
+ICOMMAND(0, clearconsole, "i", (int *type), clear_console(clamp(*type, -1, CON_MAX-1)));
+ICOMMAND(0, clearchat, "", (), clear_console(CON_MESG));
 
 void writebinds(stream *f)
 {
--- a/src/engine/main.cpp
+++ b/src/engine/main.cpp
@@ -96,7 +96,7 @@ void cleanup()
     UI::cleanup();
     cleanupwind();
     extern void clear_command(); clear_command();
-    extern void clear_console(); clear_console();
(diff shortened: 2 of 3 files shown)


src/engine/console.cpp
```diff
@@ -713,11 +713,18 @@ void processkey(int code, bool isdown)
...
 
-void clear_console()
+void clear_binds()
 {
     keyms.clear();
 }
 
+void clear_console(int type)
+{
+   if (type >= 0) { conlines[type].clear(); } else { loopi(CON_MAX) conlines[i].clear(); }
```
qreeves, line 723: I'll have to get pedantic about code formatting here.. lol.
```C++
if(type >= 0 && type < CON_MAX) conlines[type].clear();
else loopi(CON_MAX) conlines[i].clear();
```

src/engine/main.cpp
```diff
@@ -96,8 +96,9 @@ void cleanup()
     UI::cleanup();
     cleanupwind();
     extern void clear_command(); clear_command();
-    extern void clear_console(); clear_console();
+    extern void clear_binds(); clear_binds();
     extern void clear_models();  clear_models();
+    extern void clear_console(int);  clear_console(CON_MAX);
```
qreeves, line 101: I'm surprised this didn't make you segfault, as you had no guard on the range for `clear_console()` and this line here goes outside the array bounds for `conlines` (remember, `CON_MAX` may be the number of arrays, but C zero-indexes so the last one is `CON_MAX-1`). That being said, you wanted `-1` anyway to clear all console lines.

All in all, we don't really need to include `clear_console()` anyway as this is not something we need to do during `cleanup()`.

Jigoku: I just force pushed, and clamped it. Hopefully that change works out.

Jigoku: Or would it be better to clamp here:
```c
ICOMMAND(0, clearconsole, "i", (int *type), clear_console(clamp(*type, -1, CON_MAX-1)));
```

Jigoku: Looking at other ICOMMAND's i suppose it's better to clamp there. I just altered it, hopefully that suffices?

Jigoku: Only additional thing, is that it could be worth setting a minimum width value for the console UI after clearing, maybe it should be stretched across the full console width?

For example:

2020-07-07-040054_1920x1080_scrot

PR #1074, opened 2020-07-07
fix bone weight accumulation

q009:


lsalzman:
Doing this in the shader like this is expensive and not the right way to go about it. We flip the signs for the bones before we upload this to avoid this ever happening in the first place. I'd need to see a test case of where this is going wrong.

PR #1077, opened 2020-07-12
fix rollsky/rollclouds

Jigoku:
Fixes maps with yawsky set to non zero.


qreeves:
I didn't even notice that 🤦

PR #1081, opened 2020-07-23
add password prompt for private servers

Jigoku:
- add password prompt/input when clicking a server that is set to private

2020-07-23-195311_1920x1080_scrot

### DIFF
--- a/config/ui/servers.cfg
+++ b/config/ui/servers.cfg
@@ -20,6 +20,34 @@ uimenu "serversset" "Server Browser Settings" $questiontex [
     ]
 ] [] [] [] 1 []
 
+ui_serversprivate_desc = ""
+ui_serversprivate_name = ""
+ui_serversprivate_port = ""
+ui_serversprivate_pass = ""
+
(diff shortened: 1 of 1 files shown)


qreeves:
Might be a good idea to show information about the server that's asking for a password, in case of accidental clicks, or any "WTF" moments. Maybe make it clear that it wants a password to connect because `serverpass` is set, but you can also use `adminpass` to get in.

qreeves:
Bump.

Jigoku:
Still unsure of the best approach to do this.

Is that fine to use a bunch of empty vars (to stop the console errors), passed to the other function as for:
```
ui_server_private_name = ""
ui_server_private_port = ""
ui_server_private_password = ""
```

Or whether there's a better way to do it.

Jigoku:
Also, whether `adminpass` should be explicitly set/override, each time the server password is entered?

qreeves:
Initialising those variables like that is normal. `adminpass` is only for the server, not the client.

qreeves:
If that incorrect password still letting you into servers is still a thing, can you please make an issue for it?

PR #1083, opened 2020-07-28
Display IDF_CLIENT variables as "Game" in the console usage display.

shacknetisp:
This PR changes it from "Client", which is somewhat misleading, as changing the variable does not affect only your client, but the whole game.
"Game" was also the description in 1.x versions.

### DIFF
--- a/config/ui/console.cfg
+++ b/config/ui/console.cfg
@@ -130,7 +130,9 @@ ui_console_usage_text = [
     idtype_s = ""
     // IDF_CLIENT || IDF_SERVER
     if (|| (& $idflags $idfbitclient) (& $idflags $idfbitserver)) [
-        idtype_s = (concatif $idtype_s (? (& $idflags $idfbitclient) "Client" "Server"))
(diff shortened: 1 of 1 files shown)


config/ui/console.cfg
```diff
@@ -130,7 +130,7 @@ ui_console_usage_text = [
     idtype_s = ""
     // IDF_CLIENT || IDF_SERVER
     if (|| (& $idflags $idfbitclient) (& $idflags $idfbitserver)) [
-        idtype_s = (concatif $idtype_s (? (& $idflags $idfbitclient) "Client" "Server"))
+        idtype_s = (concatif $idtype_s (? (& $idflags $idfbitclient) "Game" "Server"))
```
Calinou, line 133: It might be worth adding a comment to explain why it's called "Game" instead of "Client". Just a thought, as it might not be immediately obvious at first.

shacknetisp: Good thought.

PR #1102, opened 2020-09-30
fixes #314 - add names to routes

### LINKED ISSUE #314: Give routes in race names
Names can be as discussed in #310, or some other (better) implementation. 

Names will allow us to eventually build a more robust gui for showing map routes.

Jigoku:
Fixes #314 

- `showroutenames` (toggle particle display of name)
- `routenameblend` (blend value for particle text)
- `routenames` (list of route names)

eg; map var: `routenames easy medium hard`
then becomes:

* routeid 0  = easy
* routeid 1  = medium
* routeid 2  = hard

This would now allow for a UI to choose the route by a proper name. 

One thing i'm not sure of (couldn't figure it out), is whether the particle text should only be drawn on the first node/link for the `showroutenames`, but here it just draws above each node.

2020-09-30-120511_1920x1080_scrot
2020-09-30-120512_1920x1080_scrot

### DIFF
--- a/src/game/entities.cpp
+++ b/src/game/entities.cpp
@@ -49,6 +49,9 @@ namespace entities
     VAR(IDF_HEX, routecolour, 0, 0xFF22FF, 0xFFFFFF);
     VAR(0, droproutedist, 1, 16, VAR_MAX);
     VAR(0, routemaxdist, 0, 64, VAR_MAX);
+    VAR(IDF_PERSIST, showroutenames, 0, 1, 1);
+    FVAR(IDF_PERSIST, routenameblend, 0, 1, 1);
+    SVARF(IDF_WORLD, routenames, "Easy Medium Hard", { string s; if(filterstring(s, routenames)) setsvar("routenames", s, false); });
 
     struct rail
     {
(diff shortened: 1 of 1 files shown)


qreeves:
Maybe we should just display it on the first one and show the currently selected route on the HUD? I'd like to make sure the words are capitalised when appropriate as well.

Jigoku:
Yeah, looks a bit messy with the route name spammed all over the place. I did spend a while trying to find a way to limit drawing to only the first ent within the routeid, but had no luck figuring it out.

src/game/entities.cpp
```diff
@@ -2430,6 +2433,19 @@ namespace entities
...
+    int getfirstroute()
+    {
+        if(lastroutenode == routeid) return -1;
+
+        int firstroute = -1;
+        loopenti(ROUTE)
+        if(entities::ents[i]->type == ROUTE && entities::ents[i]->attrs[0] == routeid)
+        {
+            firstroute = i; break;
+        }
```
qreeves, line 2446: Might get you to fix the formatting here first, either moving the if up on to the loop line or indenting the if block, and putting a new line before the break.

qreeves (changes requested):

src/game/entities.cpp
Jigoku: Done

PR #1109, opened 2020-10-24
fix #315 - add custom colours to routes

### LINKED ISSUE #315: Give routes in race colours
Colours can probably be implemented the same way as colours (#314).

Colours gives quick feed back to the user about which route is harder, also opens up the possibility, if someone wanted it, to render all routes at once without it becoming confusing.

The general trend seems to be the higher the number the more difficult/advanced the route is, it could be a simple interpolation from the highest route and 0 through colours (like green to red). Or we could give the ability to pick colours.

Jigoku:
Fixes #315 

- use string value for list of `routecolours` (same method as `routenames`) so that index matches `routeid`
- add usage for `routenames` and `routecolours`

Default colours are "0x00FF00 0xFF7700 0xFF0000"  (green, orange red), but can be altered on a per map basis this way.

2020-10-24-034338_1920x1080_scrot
2020-10-24-034344_1920x1080_scrot
2020-10-24-034349_1920x1080_scrot

### DIFF
--- a/src/game/entities.cpp
+++ b/src/game/entities.cpp
@@ -41,12 +41,12 @@ namespace entities
 
     VARF(0, routeid, -1, -1, VAR_MAX, lastroutenode = -1; lastroutetime = 0; airnodes.setsize(0)); // selected route in race
     VARF(0, droproute, 0, 0, 1, lastroutenode = -1; lastroutetime = 0; airnodes.setsize(0); if(routeid < 0) routeid = 0);
-    VAR(IDF_HEX, routecolour, 0, 0xFF22FF, 0xFFFFFF);
     VAR(0, droproutedist, 1, 16, VAR_MAX);
     VAR(0, routemaxdist, 0, 64, VAR_MAX);
     VAR(IDF_PERSIST, showroutenames, 0, 1, 1);
     FVAR(IDF_PERSIST, routenameblend, 0, 1, 1);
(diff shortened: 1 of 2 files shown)


src/game/entities.cpp
```diff
@@ -3006,7 +3006,17 @@ namespace entities
...
+                    char *colour = indexlist(routecolours, routeid);
+                    if(colour)
+                    {
+                        if(*colour)
+                        {
+                            part_flare(o, ents[e.links[i]]->o, 1, PART_LIGHTNING_FLARE, parseint(colour));
+                        }
+                        delete[] colour;
+                    }
+                }
```
qreeves, line 3019: The following will fix instances where the `routeid` has no matching token in `routecolours`:
```C++
int col = 0xFF22FF;
char *rcol = indexlist(routecolours, routeid);
if(rcol)
{
    if(*rcol) col = parseint(rcol);
    delete[] rcol;
}
part_flare(o, ents[e.links[i]]->o, 1, PART_LIGHTNING_FLARE, col);
```

qreeves (changes requested):
Changes needed to ensure routes render regardless of a colour being present in `routecolours`.

src/game/entities.cpp
Jigoku: done :+1:

PR #1232, opened 2021-01-28
Change specular calculation to look more realistic

10010101001:

### DIFF
--- a/config/glsl/deferred.cfg
+++ b/config/glsl/deferred.cfg
@@ -93,11 +93,9 @@ unpacknorm = [
 unpackspec = [
     result [
         vec3 camdir = normalize(camera - pos.xyz);
-        float facing = 2.0*dot(normal.xyz, camdir);
-        float specscale = min(3.0*(diffuse.a + 0.5/255.0), 2.999), gloss = floor(specscale);
-        specscale -= gloss;
-        specscale = (0.35 + 0.15*gloss) * specscale / (1.5 - specscale);
-        gloss = 5.0 + 17.0*gloss;
+        float roughness = 1.0-diffuse.a;
+		float rr = roughness*roughness;
+		float rrrr = rr*rr;
     ]
 ]
 
@@ -436,7 +434,12 @@ deferredlightvariantshader = [
                         light += diffuse.rgb*sunfacing * sunlightcolor * sunoccluded;
                     ]] [result [
                         @(if (= (+ $numlights (dlopt "c")) 1) [unpackspec])
-                        float sunspec = pow(clamp(sunfacing*facing - dot(camdir, sunlightdir), 0.0, 1.0), gloss) * specscale;
+		                vec3 C = normalize(sunlightdir+camdir);
+		                float dotCN = clamp(dot(C,normal.xyz), 0.0, 1.0);
+		                float D = ( dotCN * dotCN ) * ( rrrr - 1.0 ) + 1.0;
+		                float ldotH = clamp( dot( sunlightdir, C ), 0.0, 1.0 );
+						float VFapprox = ( ldotH * ldotH ) * ( roughness + 0.5 );
+						float sunspec = clamp(( rrrr / ( 12.5663706144 * D * D * VFapprox ) ) * sunfacing * 0.4,0.0, 9999.0);
                         @(if (dlopt "r") [result [
                             sunlight += (diffuse.rgb*sunfacing + sunspec) * sunoccluded;
                         ]] [result [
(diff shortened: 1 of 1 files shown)


qreeves:
RE coding convention requires 4 spaces instead of TAB characters. Please also provide more information in the description field of your pull requests so we better understand what it is you want to achieve and why.

config/glsl/deferred.cfg
```diff
@@ -93,11 +93,9 @@ unpacknorm = [
 unpackspec = [
     result [
         vec3 camdir = normalize(camera - pos.xyz);
-        float facing = 2.0*dot(normal.xyz, camdir);
-        float specscale = min(3.0*(diffuse.a + 0.5/255.0), 2.999), gloss = floor(specscale);
-        specscale -= gloss;
-        specscale = (0.35 + 0.15*gloss) * specscale / (1.5 - specscale);
-        gloss = 5.0 + 17.0*gloss;
+        float roughness = 1.0-diffuse.a;
```
q009, line 96: Should also be modulated by the gloss parameter, I believe.

config/glsl/ui.cfg
```diff
@@ -77,7 +82,12 @@ lazyshader 0 "modelpreview" [
                     pos.xyz /= pos.w;
                 ])
                 @(unpackspec)
-                float exspec = pow(clamp(exfacing*facing - dot(camdir, exlightdir), 0.0, 1.0), gloss) * specscale;
+                vec3 C = normalize(exdir+camdir);
```
q009, line 85: Shouldn't be exlightdir?
There's no exdir

q009 (changes requested):
Couple things to address above.

config/glsl/deferred.cfg
```diff
@@ -515,7 +519,12 @@ deferredlightvariantshader = [
                             light += diffuse.rgb*light@[j]facing * lightcolor[@@j].rgb * light@[j]shadow;
                         ]] [result [
                             @(if (= (+ $numlights (dlopt "c")) 1) [unpackspec])
-                            float light@[j]spec = pow(clamp(light@[j]facing*facing - light@[j]invdist*dot(camdir, light@[j]dir), 0.0, 1.0), gloss) * specscale;
+                            vec3 C = normalize(light@[j]dir2+camdir);
+                            float dotCN = clamp(dot(C,normal.xyz), 0.0, 1.0);
+                            float D = ( dotCN * dotCN ) * ( rrrr - 1.0 ) + 1.0;
+                            float ldotH = clamp( dot( light@[j]dir2, C ), 0.0, 1.0 );
+                            float VFapprox = ( ldotH * ldotH ) * ( roughness + 0.5 );
+                            loat light@[j]spec = clamp(( rrrr / ( 12.5663706144 * D * D * VFapprox ) ) * light@[j]facing * 0.1, 0.0, 9999.0);
```
Jigoku, line 527: There's an "f" missing on this float value which breaks things.

10010101001: Fixed

config/glsl/deferred.cfg
10010101001: diffuse.a is gloss AFAIK

q009: Yeah, you're right. diffuse.a is used for specular reflections on many textures. Gloss was just a uniform shader parameter we had to modulate roughness.

10010101001: How do I do that?

q009: We had gloss working on 0 to 2 range, 1 being the default.
So, I suppose you could just multiply diffuse.a by gloss.
Something like this
`float roughness = 1.0 - min(diffuse.a*gloss, 1);`

10010101001: The gloss you're talking about seems to be in world.cfg, gloss in deferred.cfg is defined by diffuse.a?

q009: Oh, yes. My apologies. You're absolutely right. Nevermind then :)

10010101001:
> Couple things to address above.
> Edit: just checked it myself, it seems generally broken. Most of the lighting is just gone.

Fixed

config/glsl/deferred.cfg
```diff
@@ -93,11 +93,9 @@ unpacknorm = [
...
     result [
         vec3 camdir = normalize(camera - pos.xyz);
-        float facing = 2.0*dot(normal.xyz, camdir);
-        float specscale = min(3.0*(diffuse.a + 0.5/255.0), 2.999), gloss = floor(specscale);
-        specscale -= gloss;
-        specscale = (0.35 + 0.15*gloss) * specscale / (1.5 - specscale);
-        gloss = 5.0 + 17.0*gloss;
+        float roughness = 1.0-diffuse.a;
+        float rr = roughness*roughness;
+        float rrrr = rr*rr;
```
IceflowRE, line 98: Please use another naming.
Also `rr` is not used anymore so why create it?
Not sure if RE has a pow function which you can use or just `roughness*roughness*roughness*roughness`

10010101001: Done.

q009: @IceflowRE this is GLSL, those variables get optimised out automatically by the compiler.

lsalzman:
I am not interested in making substantial changes to the lighting model at this time. There are already more than enough spec/gloss knobs in the engine that if used properly will give saner results than this, nor is this really up to the performance characteristics I was going for with the existing lighting model. As-is, this change doesn't respect any of those existing knobs so pretty much is broken atm.

10010101001:
Should I close the pull request?

q009:
> I am not interested in making substantial changes to the lighting model at this time. There are already more than enough spec/gloss knobs in the engine that if used properly will give saner results than this, nor is this really up to the performance characteristics I was going for with the existing lighting model. As-is, this change doesn't respect any of those existing knobs so pretty much is broken atm.

Yeah, I agree that this is quite a bit more math heavy than the previous implementation.
However, having tried the change myself, I can say that it provides consistently nicer looking results with many textures RE is using. Particularly due to much higher gloss range allowing for far sharper specular reflections on shinier surfaces.

So, as a middle ground solution, what if we put the change behind a variable? That way keep both the performant and well optimized code, as well as the slower, but prettier implementation. The user will be able to switch between them according to their needs.

qreeves:
Thanks for your interest in contributing, but as-is the pull request doesn't meet the requirements set by the lead programmer. I'm closing this for now, but if you're interested in contributing in the future, please join us on Discord and discuss your ideas before putting significant effort in to them so we can all get on the same page.

PR #1247, opened 2021-03-19
add camera and rail to listcomplete

dc-de:
Add camera ent and rail ent to listcomplete in newent, entfind and entfindinsel. Also removed unused from newent because this doesn't work.


qreeves:
Oh yeah, thanks!

PR #1267, opened 2021-05-30
Mouse acceleration

anengineer1:
Fixes #

- [description]
Due to how fast this game can be, I find Quake Live's mouse acceleration to work quite well in Red Eclipse.

### DIFF
--- a/src/engine/rendergl.cpp
+++ b/src/engine/rendergl.cpp
@@ -2572,11 +2572,8 @@ void gl_drawframe()
     if(!noview) gl_drawview();
     gl_drawhud(noview);
 
-    if(frametimer)
-    {
-        glFinish();
-        framemillis = getclockmillis() - totalmillis;
-    }
+    glFinish();
+    framemillis = getclockmillis() - totalmillis;
 }
 
 void cleanupgl()
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -320,6 +320,12 @@ namespace game
     FVAR(IDF_PERSIST, mousesensitivity, 1e-4f, 1, 10000);
     FVAR(IDF_PERSIST, zoomsensitivity, 0, 0.65f, 1000);
 
+    // Section regarding mouse acceleration
+    FVAR(IDF_PERSIST, mouseaccelerationexp, FVAR_MIN, 2, FVAR_MAX);
+    FVAR(IDF_PERSIST, mouseaccelerationfactor, FVAR_MIN, 0, FVAR_MAX);
+    FVAR(IDF_PERSIST, mouseaccelerationoffset, FVAR_MIN, 0, FVAR_MAX);
+    FVAR(IDF_PERSIST, mouseaccelerationsenscap, FVAR_MIN, 0, FVAR_MAX);
+
     VARF(IDF_PERSIST, zoomlevel, 0, 4, 10, checkzoom());
     VAR(IDF_PERSIST, zoomlevels, 1, 5, 10);
     VAR(IDF_PERSIST, zoomdefault, -1, -1, 10); // -1 = last used, else defines default level
@@ -2431,7 +2437,7 @@ namespace game
     {
         #define mousesens(a,b,c) ((float(a)/float(b))*c)
         if(mouseoverride&2 || (!mouseoverride && hud::hasinput(true)))
-        {
+	{
             cursorx = clamp(cursorx+mousesens(dx, w, mousesensitivity), 0.f, 1.f);
             cursory = clamp(cursory+mousesens(dy, h, mousesensitivity), 0.f, 1.f);
             if(!(mouseoverride&1)) return true;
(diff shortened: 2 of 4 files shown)


src/game/game.cpp
```diff
@@ -2441,13 +2447,55 @@ namespace game
             physent *d = (!gs_playing(gamestate) || player1->state >= CS_SPECTATOR) && (focus == player1 || followaim()) ? camera1 : (allowmove(player1) ? player1 : NULL);
             if(d)
             {
-                float scale = (focus == player1 && inzoom() && zoomsensitivity > 0 ? (1.f-((zoomlevel+1)/float(zoomlevels+2)))*zoomsensitivity : 1.f)*sensitivity;
+	        float accelsensitivity = 1;
```
q009, line 2450: Please fix indentation in various parts of the PR

src/engine/rendergl.cpp
```diff
@@ -2572,11 +2572,8 @@ void gl_drawframe()
     if(!noview) gl_drawview();
     gl_drawhud(noview);
 
-    if(frametimer)
-    {
-        glFinish();
-        framemillis = getclockmillis() - totalmillis;
-    }
+    glFinish();
```
q009, line 2575: What's the reason behind change here?

q009 (changes requested):

src/engine/rendergl.cpp
anengineer1: it is important for the framemillis to be calculated for the mouse acceleration to work; it is required to have the same accel no matter your framerate.

With the previous code, framemillis was only calculated when enabling the frametimer.

src/game/game.cpp
anengineer1: Sorry for my ignorance, but when checking at the file itself, it seems to be correctly indented ->

https://github.com/anengineer1/base/blob/mouse_acceleration/src/game/game.cpp

I'm having a close look at the file at the moment

q009: Make sure you're using spaces, not tabs

anengineer1: Sorry for the delay, seems all is in order now, I just deleted all the tabs and pressed "space", I'm removing the autotab indent from my editor.

q009 (approved):
LGTM

q009 (changes requested):
Actually, one more thing: squash those changes into one or two commits please. It's quite a mess

anengineer1:
You mean, the 7 or so attempts in removing the tabs? I'm learning how to do so, I'm on it

anengineer1:
I think I solved the thingy; I'm learning a lot today.

Please, tell me if it is suitable now.

q009 (approved):

src/engine/rendergl.cpp
```diff
@@ -2572,11 +2572,8 @@ void gl_drawframe()
...
+    framemillis = getclockmillis() - totalmillis;
```
qreeves, line 2576: My question here: is `framemillis` really necessary when you have `curtime`?

src/game/game.cpp
```diff
@@ -320,6 +320,12 @@ namespace game
     FVAR(IDF_PERSIST, mousesensitivity, 1e-4f, 1, 10000);
     FVAR(IDF_PERSIST, zoomsensitivity, 0, 0.65f, 1000);
 
+    // Section regarding mouse acceleration
```
qreeves, line 323: I'd prefer these variables to be a little shorter, so maybe `mouseaccel*`? Also, seeing as `mouseaccelerationfactor` is the switch variable as well, just call it plain `mouseaccel`.

src/game/game.cpp
```diff
@@ -2441,7 +2447,49 @@ namespace game
             physent *d = (!gs_playing(gamestate) || player1->state >= CS_SPECTATOR) && (focus == player1 || followaim()) ? camera1 : (allowmove(player1) ? player1 : NULL);
             if(d)
             {
-                float scale = (focus == player1 && inzoom() && zoomsensitivity > 0 ? (1.f-((zoomlevel+1)/float(zoomlevels+2)))*zoomsensitivity : 1.f)*sensitivity;
+                float accelsensitivity = 1;
+                if(mouseaccelerationfactor)
```
qreeves, line 2451: I try to make non boolean or integer variables use a comparison check rather than an implicit boolean, so I would prefer `if(mouseaccel > 0)`.

src/game/game.cpp
```diff
@@ -2441,7 +2447,49 @@ namespace game
...
+                {
+                     // Then do Quake Live-style power acceleration.
+                     // Note that this behavior REPLACES the usual
+                     // sensitivity, so we apply it but then dividie by
+                     // sensitivity.value so that the later multiplication
+                     // restores it again.
```
qreeves, line 2457: We value screen space with the code, so please try to compress these comments in to one or two lines.

src/game/game.cpp
```diff
@@ -2441,7 +2447,49 @@ namespace game
...
+                     float mouse_deltadist = sqrtf(dx * dx + dy * dy);
+                     float speed = mouse_deltadist / framemillis;
+                     float adjusted_speed_pxms = (speed * 0.001f - mouseaccelerationoffset) * mouseaccelerationfactor;
+                     float inv_sensitivity = 1.0f / sensitivity;
```
qreeves, line 2461: I'm probably being really pedantic, but again, compressed into fewer lines would be better. With them all being the same type you don't really need to declare it more than once. Also, I prefer variables without underscores as it is an awkward character to type when you're bashing out code.

src/game/game.cpp
```diff
@@ -2441,7 +2447,49 @@ namespace game
...
+                     if (adjusted_speed_pxms > 0)
```
qreeves, line 2462: Convention with the codebase is to not have a space between `if` and the opening bracket. Ditto for lines 2464 and 2483.

src/game/game.cpp
```diff
@@ -2441,7 +2447,49 @@ namespace game
...
+                     }
+                     if (mouseaccelerationsenscap > 0.0f && accelsensitivity > mouseaccelerationsenscap * inv_sensitivity)
+                     {
+                         // TODO: How does this interact with sensitivity changes? Is this intended?
+                         // Currently: senscap is in absolute sensitivity units, so if senscap < sensitivity, it overrides.
+                         accelsensitivity = mouseaccelerationsenscap * inv_sensitivity;
+                     }
+
+                }
+                // end accel
```
qreeves, line 2491: This comment isn't really necessary. I know it's a pain we don't comment a lot, and I've let the comments above slide because they're very speculative changes, but we try to make it so the code speaks for itself and rely on fewer comments to make it quicker to read through actual code.

qreeves (changes requested):
I really appreciate the contribution, and this is probably something we've needed for some time. Please just bear without our nit-picking as we try to keep our codebase fairly pure.

src/engine/rendergl.cpp
qreeves: [19:52] Q009: @Quin curtime is scaled
[19:52] Q009: So, just changing the gamespeed would have an effect on the mouse acceleration when using curtime to calculate it(edited)
[19:52] Q009: Which is definitely not what one would want

anengineer1:
Sorry for the delay, I took this oportunity to further test the code just in case, everything should be in order now.

Pls, tell me if something more is necessary.

qreeves (approved):

qreeves:
No problem, we all go at our own pace around here, so no need to rush. Thanks so much for pushing the requested changes, this is now approved for merging and will get done in my next round of merges (I'll need to test it).

src/game/game.cpp
```diff
@@ -2441,7 +2447,43 @@ namespace game
...
+                             // The limit of the then-branch for mauseaccel -> 1.
+                              accelsensitivity += (1 / sensitivity);
+                             // Note: QL had just accelsens = 1.0f.
+                             // This is mathematically wrong though.
+                         }
+                     }
+                     else
+                     {
+                         // The limit of the then-branch for adjustedspeed -> 0.
+                         accelsensitivity += 0.0f;
```
qreeves, line 2476: I merged this and took a proper look at it, and this line is pretty much a non-statement. Is it supposed to be something else?

anengineer1: You are right.

This is not the first time I implemented this mouse acceleration and the maintainer of another project prefered to have it that way to make it more explicit so I wrote the mouse acceleration in a very similar way than in that other game engine I helped with.

Feel free to change it.

anengineer1: If you want me to change it just tell me and I will gladly will.

qreeves: Nah, all good, I'll do it in my tidy up commit later. Just wanted to make sure it wasn't a typo.

q009:
Due to performance issues, I had to revert the change.
The ``glFinish`` call in ``gl_drawframe`` drops the framerate by about 30%

anengineer1:
Thank you a lot for letting me know!

Once I have time again, I will work on it.

Do you think the cause might be using the frametime counter? it is very important for making the accel frame-independent.

Tesseract game has a mouse acceleration implementation, I will take a look at its source code and see what they have done

Sorry for taking so long in answering.

q009:
The cause is using relying on ``glFinish``. It stalls the entire pipeline until the processing on the GPU is finished, drastically reducing the performance. Perhaps try using ``curtime`` instead to base the acceleration on, or ``elapsedmillis``.

anengineer1:
Thank you a lot! I'll take a look at it!

I'll will be very absent during this month but I will get my hands dirty as soon as I can

anengineer1:
Greetings, I've made some testing using curtime instead of framemillis and it seems it works as planned:

- Acceleration curves are consistent not mattering the framerate
- I also tested the accel on high graphics to force inconsitent fluctuation and the acceleration curve is also consistent and aiming it's still a pleasure.

My branch is updated but I don't know how to update the thing on the pull request.


Sorry and thank you

PR #1308, opened 2021-11-16
add ui setting Show Clips

dc-de:
add ui setting Show Clips to enable and disable showclips.

### DIFF
--- a/config/ui/settings.cfg
+++ b/config/ui/settings.cfg
@@ -1146,13 +1146,13 @@ uimenu "settings" "Settings" "textures/icons/settings" [
                                 ]
                                 uitablerow [
                                     uihlist $ui_padbutton [
-                                      uicheckbox "Show Engine Timings" $timer $ui_checksize [timer = (! $timer)] [] [] [] [] "settings_timer"
+                                      uicheckbox "Show Clips" $showclips $ui_checksize [showclips = (! $showclips)] [] [] [] [] "settings_showclips"
                                       uialign -1 0
                                     ]
                                 ]
                                 uitablerow [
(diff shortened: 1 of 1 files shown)


config/ui/settings.cfg
```diff
@@ -1150,6 +1150,12 @@ uimenu "settings" "Settings" "textures/icons/settings" [
                                       uialign -1 0
                                     ]
                                 ]
+                                uitablerow [
+                                    uihlist $ui_padbutton [
+                                      uicheckbox "Show Clips" $showclips $ui_checksize [showclips = (! $showclips)] [] [] [] [] "settings_showclips"
+                                      uialign -1 0
+                                    ]
+                                ]
```
Calinou, line 1158: Should the setting be moved higher up? I'd put it below "Show Speed" so that "Show Engine Timings" remains at the bottom.

dc-de: I am not sure how far we would change everything in the UI arround so maybe the order would be changed again and again.

dc-de: @Calinou done now the order is better.

Jigoku:
I wonder if it may be better the add a crosshair tab to the settings menus, and put that option in there maybe? The other crosshair settings have still yet to be ported to the new UI system, but they would more than likely end up in that tab  :+1:

dc-de:
> I wonder if it may be better to add a crosshair tab to the settings menus, and put that option in there maybe? The other crosshair settings have still yet to be ported to the new UI system, but they would more than likely end up in that tab +1

yeah good idea maybe I should bundle all crosshair things together lol.

PR #1309, opened 2021-11-19
Default sunlight for newmap

Jigoku:
- add a neutral sunlight to brighten things up a bit


qreeves:
Should this be a code modification in stead? I tend to prefer to put the defaults there.

Jigoku:
> Should this be a code modification instead? I tend to prefer to put the defaults there.

Probably best yeah. I did initially add one of the cloud skybox's along with sunpitch / sunyaw etc, however the file above didn't want to set the sun position, assuming that could be done properly in the code. 

I guess i'll close this then.

PR #1314, opened 2021-12-11
add moveselents command

Jigoku:
works similar to moveblendmap / moveselwaypoints

### DIFF
--- a/src/engine/world.cpp
+++ b/src/engine/world.cpp
@@ -1493,3 +1493,10 @@ void mpeditent(int i, const vec &o, int type, attrvector &attr, bool local)
     clearshadowcache();
     commitchanges();
 }
+
+void moveselents(vec &offset)
+{
+    if(noentedit()) return;
+    groupedit(e.o.add(offset));
+}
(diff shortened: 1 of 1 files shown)


src/engine/world.cpp
```diff
@@ -1493,3 +1493,10 @@ void mpeditent(int i, const vec &o, int type, attrvector &attr, bool local)
     clearshadowcache();
     commitchanges();
 }
+
+void moveselents(vec &offset)
+{
+    if(noentedit()) return;
+    groupedit(e.o.add(offset));
+}
+ICOMMAND(0, moveselents, "fff", (float *x, float *y, float *z), vec v(*x, *y, *z); moveselents(v));
```
qreeves, line 1502: No newline at end of file.

qreeves (changes requested):
Only one issue to fix prior to merge.

src/engine/world.cpp
Jigoku: whoops, fixed.

qreeves (approved):

PR #1316, opened 2021-12-19
add listbans command

Jigoku:
Allows server moderators to list active bans within the server in-game.

- `/listbans`
- Restricts to G(banlock) privileges (but maybe this would be better as PRIV_ADMINISTRATOR ?)

I did intend to also add user tracking `(added by: person)` etc, although couldn't get ipinfo to store it properly right now. Whether that's something to add later.

2021-12-19-064527_1920x1080_scrot


Jigoku:
Now outputs the name of the person who applied the ban; 
```
1: [ban] 192.168.0.69 (reason: trolling [unixfreak])
2: [ban] 192.168.0.42 (reason: more trolling [unixfreak])
```
etc

qreeves:
- I think it is unwise to include the name of the person from the client's side, as this could be edited out by a malicious user.
- If a user is authed, it should probably include their authname.
- If you're going to add a list command for one thing, it's probably a good idea to do it for all the `addcontrol` types.
- When you change the network protocol, you need to bump `GAMEVERSION` in `src/game/game.h`.

q009:
Any updates on this?

PR #1322, opened 2021-12-28
Automatically adjust field of view for widescreen aspect ratios

Calinou:
Inspired by https://github.com/assaultcube/AC/pull/424.

The FOV value is now always interpreted as being a 4:3 horizontal FOV value. It will be automatically scaled for different aspect ratios (which results in a horizontal FOV of ~106.26 degrees on 16:9, for example).

This provides a better out of the box experience for most players, since 16:9 is the norm for gaming nowadays (or even wider on mobile).

Default FOV values were changed to be less extreme at wider aspect ratios.


src/game/game.cpp
```diff
@@ -679,12 +679,17 @@ namespace game
...
     {
-        if(player1->state == CS_EDITING) return editfov;
-        if(focus == player1 && player1->state == CS_SPECTATOR) return specfov;
-        if(thirdpersonview(true)) return thirdpersonfov;
-        return firstpersonfov;
+        float resultfov = 90.0f;
+        if(player1->state == CS_EDITING) resultfov = editfov;
+        if(focus == player1 && player1->state == CS_SPECTATOR) resultfov = specfov;
+        if(thirdpersonview(true)) resultfov = thirdpersonfov;
+        resultfov = firstpersonfov;
```
qreeves, line 688: Probably missing `else`.

qreeves (commented):
While I have pointed out an issue with the code, I don't know if this is generally a good idea.

Most people who change FOV are attempting to set specific values and this change results in behaviour which is fairly hidden from the user. Most people would probably be left wondering why the values they set aren't properly reflected.

That being said, it should be noted that FOV is already aspect-corrected in Cube 2 and family on the Y-axis, rather than the X-axis as is being done here. I don't know if this was ported to AC but I do know the change was done in the last decade or so, which is after AC forked from Cube 1.

I think if this were to be merged, it'd need to be an optional component, rather than a forced default, but I invite further discussion on the issue.

src/game/game.cpp
```diff
@@ -680,12 +680,17 @@ namespace game
...
-        return firstpersonfov;
+        float resultfov = 90.0f;
+        if(player1->state == CS_EDITING) resultfov = editfov;
+        if(focus == player1 && player1->state == CS_SPECTATOR) resultfov = specfov;
+        if(thirdpersonview(true)) resultfov = thirdpersonfov;
+        else resultfov = firstpersonfov;
+
+        // automatic FOV adjustment for wide screens
+        // (horizontal FOV value is always specified for 4:3, but scaled for different aspect ratios)
+        return atan(tan(resultfov * M_PI / 360.0f) * 0.75f * aspect) * 360.0f / M_PI;
```
q009, line 693: Since this is a change that drastically changes the expected behavior of the FOV math, this should be behind an additional variable that enables/disables this. Gate it behind "fovaltaspectcorrect" variable and I think it'll be good to merge.

q009 (changes requested):

PR #1323, opened 2021-12-28
Add variables for constant first-person model offset

Calinou:
This can be used to make the first-person weapon model less intrusive (or left-handed, although this isn't ideal as the model itself won't be mirrored).

## Preview

Various combinations of `/firstpersonside`, `/firstpersonup` and `/firstpersondist`:


SniperGoth:
Oh this could be useful, without having to manually change the offset values on the cfg files.

Calinou:
> Oh this could be useful, without having to manually change the offset values on the cfg files.

Indeed, this is also useful for people working on the models themselves :slightly_smiling_face: 

Eventually, we can explore adding a variable for a proper left-handed mode that will flip the weapon models.

Calinou:
Superseded by https://github.com/redeclipse/base/pull/1324 (wrong fork branch).

PR #1324, opened 2021-12-28
Add variables for constant first-person model offset

Calinou:


src/game/game.cpp
```diff
@@ -3745,6 +3749,18 @@ namespace game
...
+
+                vec dirforward = vec(mdl.yaw*RAD, mdl.pitch*RAD);
+                vec dirside = vec((mdl.yaw+90)*RAD, 0.0f);
+                vec dirup = vec(mdl.yaw*RAD, (mdl.pitch+90)*RAD);
+
+                vec trans = vec(0, 0, 0);
+                trans.add(vec(dirforward).mul(firstpersondist));
+                trans.add(vec(dirside).mul(firstpersonside));
+                trans.add(vec(dirup).mul(firstpersonup));
+                mdl.o.add(trans);
```
qreeves, line 3762: There's not much point to any of the intermediate `vec`'s here.

qreeves (commented):
My only concern here is that the first person arms determine the muzzle location, and it would be possible to set them to a position that would pretty much allow a person to shoot around corners. The position may need to be constrained to the player's width, but I'm not certain if that is enough to mitigate the issue.

q009:
Any updates on this?

PR #1325, opened 2021-12-28
Tweak the default `filetimeformat` to be more readable

Calinou:
For example, `20211228224900` will now be `20211228_224900`.

This mainly affects saved screenshots and demos.


qreeves:
I don't particularly like hyphens in file names, but this is more of a personal opinion than a genuine criticism. What I am concerned about is how this increases the length of the filenames considerably, which may have impacts further down the line where they need to be displayed (eg. menus) and is the reason I originally picked such an obtuse formatting arrangement. I think a better compromise might be `%Y%m%d_%H%M%S`, like that used by most cameras/smartphones. Further discussion may be warranted.

Calinou:
> I think a better compromise might be `%Y%m%d_%H%M%S`, like that used by most cameras/smartphones. Further discussion may be warranted.

Done, I changed it to `%Y%m%d_%H%M%S`.

PR #1326, opened 2021-12-28
Improve camera smoothing when crouching

Calinou:
Asymptoptic (ease-out) smoothing is now used instead of linear interpolation. This kind of smoothing is often used for camera crouching animations in modern FPS games (see Doom 3 or Xonotic for examples).

This smoothing has been tested to be FPS-independent. It works well in all common situations (crouching/uncrouching, moving while crouching, crouchsliding).


src/game/game.cpp
```diff
@@ -1275,15 +1275,18 @@ namespace game
...
                 {
+                    // asymptoptic smoothing of crouch view height (smoother than linear interpolation)
+                    float zamt = abs(d->height - zrad) * 0.35f * zoff*curtime/float(PHYSMILLIS);
                     if(d->actiontime[AC_CROUCH] <= 0) d->actiontime[AC_CROUCH] = lastmillis;
                     if(d->height > zrad && ((d->height -= zamt) < zrad)) d->height = zrad;
                     else if(d->height < zrad && ((d->height += zamt) > zrad)) d->height = zrad;
                 }
                 else
                 {
+                    // asymptoptic smoothing of crouch view height (smoother than linear interpolation)
```
qreeves, line 1288: In a codebase fairly devoid of comments where concise code is prioritized, having the same comment twice that takes up its own line is kind of a waste of space.

qreeves (commented):
I don't know if I understand the point of this, the demonstration isn't really clear on what is going on, and I'm looking at this `0.35` and wondering what the hell it is for. Maybe I just haven't had enough coffee yet.

Looking at the cheat sheet provided, it makes me wonder if we should be using a more elastic easing method.

Calinou:
> and I'm looking at this 0.35 and wondering what the hell it is for.

The `0.35f` is a magic value that scales the animation speed (higher is faster). I should have made this clearer in the code :slightly_smiling_face:

qreeves:
> The `0.35f` is a magic value that scales the animation speed (higher is faster). I should have made this clearer in the code 🙂

Should probably be a variable then 👍

q009:
What's the status on this PR? I see no problems with this, other than the previously mentioned issue not being addressed.
Once there's a game variable for the easing of the animation, I think it'll be good to merge.

src/game/game.cpp
```diff
@@ -1276,15 +1276,19 @@ namespace game
             }
             if(crouching || d->crouching(true))
             {
-                float zamt = zoff*curtime/float(PHYSMILLIS);
+                float crouchanimspeedscale = 0.35f;
```
q009, line 1279: Make this a game variable as requested

q009 (changes requested):

PR #1330, opened 2021-12-30
Use in-out easing for better animation interpolation

Calinou:
Follow-up to https://github.com/redeclipse/base/pull/1326.

The difference is mainly noticeable on high refresh-rate monitors, but it affects both third-person animations and first-person weapons. The effect on first-person weapons is mainly visible after switching to certain weapons such as the pistol, shotgun and zapper.

The animation interpolation duration has been slightly lengthened to better make use of the new easing.

### DIFF
--- a/src/engine/animmodel.h
+++ b/src/engine/animmodel.h
@@ -1123,7 +1123,9 @@ struct animmodel : model
                     if(diff<aitime)
                     {
                         p.prev.setframes(d->animinterp[interp].prev);
-                        p.interp = diff/float(aitime);
+                        // use easeInOutSine easing for smoother appearance
+                        // https://easings.net/#easeInOutSine
+                        p.interp = -(cos(M_PI * diff/float(aitime)) - 1) / 2;
                     }
                 }
             }
--- a/src/engine/rendermodel.cpp
+++ b/src/engine/rendermodel.cpp
@@ -1,7 +1,7 @@
 #include "engine.h"
 
 VAR(0, oqdynent, 0, 1, 1);
-VAR(0, animationinterpolationtime, 0, 200, 1000);
+VAR(0, animationinterpolationtime, 0, 220, 1000);
 
 model *loadingmodel = NULL;
 


qreeves:
Discord copypasta:

> [09:22] Q009: Calinou: isn't that going to change the pacing between animation frames? 
> [09:23] Q009: Turning a motion that was supposed to be linear and steady, into one that's constantly easing in-out with each frame 
> [09:23] Calinou: that might sound like it, but it works great in practice
> [09:24] Q009: Yeah, that really doesn't sound right
> [09:25] Calinou: I'm adding comparison videos
> [09:25] Q009: Because it has a literal potential to cause weird animations
> [09:25] Q009: Even if you don't see anything wrong right now
> [09:25] Q009: There might be some cases that looks weird 
> [09:26] Calinou: I've tried various cases (jigglestrafing on ground and air, crouch-spamming, etc) and it didn't break so far
> [09:26] Q009: No, not cases like that 
> [09:26] Q009: But cases with linear motion 
> [09:26] Q009: e.g. a box going slowly straight, at a steady pace 
> [09:27] Q009: That will surely be fucked up with this change
> [09:27] Q009: Because it will no longer be going at a steady pace
> [09:27] Q009: If anything it'd be better to have a choice for the interp type, per-model
> [09:28] Q009: Linear, or sine
> [09:28] Calinou: I'm not sure if we have such animations in the game right now
> [09:28] Q009: Can't login on GH right now, which is why I'm posting it here
> [09:28] Q009: We do
> [09:28] Q009: There's at least one
> [09:29] Q009: It's in the process of being scrapped though
> [09:29] Q009: But still
> [09:29] Q009: Code that actually breaks animations by default is a big no no from me
> [09:30] Q009: mdlinterp though, would be fine with me
> [09:30] Q009: 0 - no interp, 1 - linear, 2 - sine
> [09:31] Q009: And then, all organic (non-mechanical) animations could use sine interpolation
> [09:33] Q009: Though, the overall visual benefit is still probably extremely limited and subtle
> [09:33] Q009: Given the complication that this introduces, it's not ideal :P

q009:
Yeah, as in the pasted discussion, I'm not a fan of this idea at all. However, I agree with the fact that animations could be done better. Instead of doing this in the code, I plan on remaking some of the animations, particularly the weapon switch ones, to make them look smoother and more natural, as part of the ongoing process.

PR #1333, opened 2022-01-04
Mouse acceleration

anengineer1:
Fixes #

- Due to how fast this game can be, I find Quake Live's mouse acceleration to work quite well in Red Eclipse.

This is has been previously discussed here:
- https://github.com/redeclipse/base/pull/1267


q009:
If this version doesn't suffer from the previous problems, I see no reason not merge it.
Will need some brief testing...

anengineer1:
Greetings, I made this like 3 years ago, when I tested it after following your advice I didn't notice the issues you mentioned in the previous attempt... can someone test it?

PR #1339, opened 2022-02-19
clipfacevecs negative offset fix proposal

devnexen:
- clipfacevecs: Fixing crash on party launch when negative offset access attempts


Calinou:
Could you give some more context to reproduce the crash? Reading the PR description, I don't understand what's going on.

devnexen:
After choosing a server and before the party begins I encounter a crash and looking at the root issue leads to negative access to the arrays aforementioned.

qreeves:
This is not the correct solution to a problem of this kind. Please provide a backtrace of the crash so the originating function can be identified.

qreeves:
More than one month without a response to a PR comment, closing.

PR #1343, opened 2022-04-24
fix msaa on AA ultra settings

dc-de:
msaa 16 and everything grater msaa 1 messed up the flame particles and turn them into kinda black haze
only msaa 0 and msaa 1 works
I am still not sure if msaa 0 or msaa 1 fits better for the ultra settings

### DIFF
--- a/config/ui/settings.cfg
+++ b/config/ui/settings.cfg
@@ -317,7 +317,7 @@ aalevels = [
         tqaa 0
         smaa 0
         smaaquality 0
-        msaa 16
+        msaa 0
     ]
 ]
 //ui scales by area (2x area = 1.4x linear dimension increase)


Calinou:
MSAA is expensive in Tesseract, so I recommend not relying on it and using SMAA + TAA if you want the highest quality antialiasing.

If you find that TAA adds too much blurriness:

- Make sure your FPS is not limited by V-Sync or `/maxfps`, and try to get as much FPS as you can by reducing graphics settings. This will allow TAA to resolve faster, reducing ghosting artifacts and improving antialiasing quality in moving scenes.
- Support for supersampling could be added by allowing `/gscale` values above 100 (up to 200 for 4× SSAA).
- Contrast-adaptive sharpening could be added. It's a single-pass shader, so it's as easy to integrate as SMAA.

qreeves:
Fixed upstream

dc-de:
8× MSAA is not working right for the new flame particles
as I mention before only 0 and 1 MSAA works normal for these particles

PR #1384, opened 2022-10-31
SDL.h to SDL2/SDL.h (compiles on SDL1 systems).

Sondro:
Fixes #
Compile will fail on systems that have SDL1 installed (which is still updated for older apps like Cube1 -- this version works on both & should be SDLx future proof).

SDL.h to SDL2/SDL.h

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -393,14 +393,6 @@ clean-tessfont:
 
 clean: clean-enet clean-client clean-server clean-genkey clean-tessfont
 
-
-DEPENDS  = $(shell find . -name '*.d')
-ALL_OBJS = $(shell find . -name '*.o')
-
-clean2:
-	@echo "'clean' = remove all *.d & *.o in project:" 
-	@rm -rf $(DEPENDS) $(ALL_OBJS)
-
 %.h.gch: %.h
 	$(CXX) $(CXXFLAGS) -x c++-header -o $(subst .h.gch,.tmp.h.gch,$@) $(subst .h.gch,.h,$@)
 	$(MV) $(subst .h.gch,.tmp.h.gch,$@) $@
--- a/src/Makefile
+++ b/src/Makefile
@@ -103,7 +103,7 @@ ifneq (,$(findstring msvc,$(PLATFORM)))
     CC = clang
 else
     override CXX := $(TOOLSET_PREFIX)$(CXX)
-    override CC := $(TOOLSET_PREFIX)$(CC)
+    override CC  := $(TOOLSET_PREFIX)$(CC)
 endif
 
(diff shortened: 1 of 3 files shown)


qreeves:
Does it actually fail though? RE bundles it's own version of the SDL headers.

Sondro:
Yeah, it says: fatal error:SDL.h: No such file or directory #include <SDL.h>

qreeves:
Ah alright, I'll have to look into this further when I'm at my computer as there might be other things that need tweaking after this change.

Hirato:
I'm confused, how does it attempt to compile with SDL1 at all?
The build commands invokes `pkg-config --cflags x11 sdl2 SDL2_image openal sndfile zlib gl` which adds (among others) `-I/usr/include/SDL2`, meaning `#include <SDL.h>` should resolve to `/usr/include/SDL2/SDL.h`

All the SDL1 headers are in `/usr/include/SDL`; how is that SDL.h making it into your build at all?

Sondro:
I solved the problem without stackoverflow, but you can see a similar problem & solutions posted there: https://stackoverflow.com/questions/10488775/sdl-h-no-such-file-or-directory-found-when-compiling

Hirato:
As several of the comments there explain, one way to include headers is to tell the compiler WHERE to find the headers via one of 3 methods:
1. adding an explicit `-I/usr/include/SDL2` directive
2. using `pkg-config --cflags sdl2` to do that
3. or using `sdl2-config --cflags` to do that.

We are using the 2nd method, but we're also looking up `SDL2_image`; between that and `sdl2`, at least one of them should add the proper search path.
If it's genuinely not finding the proper SDL header for you, then something is uniquely wrong with either your system, your distro, or your install of SDLs 1 and 2.
IMO, You'd be better served in finding out why the pkg-config command isn't including the correct search path via cflags, rather than changing the code to include the header relative to `/usr/include/`.

qreeves:
> As several of the comments there explain, one way to include headers is to tell the compiler WHERE to find the headers via one of 3 methods:
> 
> 1. adding an explicit `-I/usr/include/SDL2` directive
> 2. using `pkg-config --cflags sdl2` to do that
> 3. or using `sdl2-config --cflags` to do that.
> 
> We are using the 2nd method, but we're also looking up `SDL2_image`; between that and `sdl2`, at least one of them should add the proper search path. If it's genuinely not finding the proper SDL header for you, then something is uniquely wrong with either your system, your distro, or your install of SDLs 1 and 2. IMO, You'd be better served in finding out why the pkg-config command isn't including the correct search path via cflags, rather than changing the code to include the header relative to `/usr/include/`.

This is the accepted answer.

PR #1394, opened 2022-12-25
`tools.h` vector object commenting

no-lex:
- Documents the member functions in `vector<class T>` that are common to Libprimis' former implementation.

The Imprimis project no longer has any use for the `vector` documentation it had written, since it has been replaced with `std::vector` in all instances. Because of this, many entries describe comparisons to `std::vector`.

All of the comments are in Doxygen format.

### DIFF
--- a/src/shared/tools.h
+++ b/src/shared/tools.h
@@ -793,20 +793,52 @@ template <class T> struct vector
 {
     static const int MINSIZE = 8;
 
-    T *buf;
-    int alen, ulen;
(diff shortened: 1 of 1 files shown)


qreeves:
I really appreciate the effort and helpful intentions here, but the internal discussion has resulted in the decision to retain readable code over verbose commenting as per Cube 2 style guidelines. Doxygen comments on every member of a class is just too much.

I hope Imprimis is going well and that you're flourishing in an environment with less constraints on what you can do and accomplish. Let me know if you ever need anything.

PR #1409, opened 2023-02-13
Update version.cfg

dc-redeclipse:
Fix map name from parker to park
Add capture (ctf) to cargo16
Add new mostly AI generated sci-fi theme fitting names for bots


config/version.cfg
```diff
@@ -26,60 +26,178 @@ version_maplist = [
...
+    Foster
+    Fox
+    Ghost
+    Gideon
+    Harper
+    Havoc
+    Hayden
+    Hunter
+    Ian
+    Icon
```
IceflowRE, line 57: Would exclude this one.

config/version.cfg
```diff
@@ -26,60 +26,178 @@ version_maplist = [
...
+    Naya
+    Nia
+    Ophelia
+    Petra
+    Phoenix
+    Polaris
+    Qiana
+    Rebekah
+    Rhea
+    Riot
```
IceflowRE, line 168: Should be fine as itself, but probably remove to due to name clash with Riot Games.

qreeves:
Okay, so my problems with this are:
- There are far fewer names than there used to be, this will result in much less variety.
- The lines should be combined as much as possible, like how it was, otherwise the list takes up way too much space and makes the file difficult to navigate.

config/version.cfg
IceflowRE: Thats true, just wanted to point that out.

dc-de: Sure... good note! The AI is not perfect for sure.

dc-de:
It should be now solved.
The AI is stupid, won't give me more useful names, but I think we have enough now.

dc-de:
Fixed @qreeves
sorry, I forget to resolve this

dc-redeclipse (commented):
more sci-fi names

dc-redeclipse:
@qreeves the names should be now fine.
I add a bunch more fitting the sci-fi theme and cyber look.
Is it now ok?

qreeves:
Yeah, but you'll need to rebase on current master as there's a conflict that's preventing merging.

PR #1412, opened 2023-02-14
Update profile.cfg

dc-de:
Unifying naming

### DIFF
--- a/config/ui/profile.cfg
+++ b/config/ui/profile.cfg
@@ -157,7 +157,7 @@ uimenu "welcome" "Welcome to Red Eclipse" $alerttex [
         uialign 0 -1
         uifill .6 0
         case $ui_profile_welcome_tab 0 [
-            uitext "Welcome to Red Eclipse ^f7."
+            uitext "Welcome to Red Eclipse."
             uispace 0 $ui_padnormal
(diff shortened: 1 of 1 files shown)


config/ui/profile.cfg
```diff
@@ -151,13 +151,13 @@ ui_profile_preview = [
...
 //step-by-step introduction to player setup
-uimenu "welcome" "Welcome to Red Eclipse 2" $alerttex [
+uimenu "welcome" "Welcome to Red Eclipse" $alerttex [
     uifill .6 .85
     uivlist $ui_padbutton [
         uialign 0 -1
         uifill .6 0
         case $ui_profile_welcome_tab 0 [
-            uitext "Welcome to Red Eclipse ^f02^f7."
+            uitext "Welcome to Red Eclipse ^f7."
```
qreeves, line 160: You can remove the "` ^f7`"

qreeves (changes requested):

dc-de (commented):
Removed double text color

config/ui/profile.cfg
dc-de: yeah sorry that is doubled, removed in patch #7c3d869

qreeves (approved):

PR #1455, opened 2024-02-18
disable taa blurring by default

10010101001:
Fixes #

- default tqaaquincunx to 0, disabling taa blurring.


Calinou:
This likely makes TAA a lot less effective, so it needs comparisons to be done at identical framerates (`/maxfps 60`). Remember that TAA is more effective at higher framerates since it can converge faster.

Tesseract's TAA is quite good at not being too blurry (it fares much better than in most other engines in this regard), so I'm not sure if we should nerf it this way.

qreeves:
It isn't clear why this change is being suggested. There are reasons behind `tqaaquincunx` defaulting to `1`.

PR #1567, opened 2024-07-10
add temporary loadout

### LINKED ISSUE #1363: Easier way to temporarily change loadout between deaths
As discussed on Discord, I feel it would be nice if players had more control over their loadout and were encouraged to switch it often based on circumstances and changes in the match. Currently you have the profile page where you can choose random weapons without a customizable pattern, and can set the page to show whenever a match starts... I still find this rather limiting however, you probably don't want to change your entire profile and only at the start of every match. Also the changes shouldn't be permanent: You can configure a default loadout as you do now, but if you pick specific weapons during a match it should be only for that match.

My suggested solution was partly inspired from Overwatch, where between deaths you can select a new hero: What if in our case we had a quick loadout selector which shows during death? Ideally it won't need to be a menu at all: Just a special HUD element showing your current loadout as the player uses shortcut keys to change the entry per slot. You could for instance have a vertical list per loadout showing all weapon icons with the active one selected: Hold down 1 and use the mouse wheel to scroll to a different weapon, 2 for the second slot, etc. If not you could have a key to toggle a quick popup for temporary loadout changes only... or even better a pie menu similar to the one for voice actions (V key by default).

nochinator:
Fixes #1363

- add new SVAR: playertemploadweap, most instances of playerloadweap now use playertemploadweap, some specificly don't (such as the reset button)
- the reset button has been renamed to "Undo Changes" for clarity
- when playerloadweap changes it also sets playertemploadweap to the same value
- playertemploadweap now also can be updated without affecting playerloadweap, with a button on the loadout menu to do so.

### DIFF
--- a/config/ui/game/hud/pie.cfg
+++ b/config/ui/game/hud/pie.cfg
@@ -440,7 +440,7 @@ game_hud_piemenu_weapsel_checkloadout = [
     local _result
     _result = 1
 
-    looplist i $playertemploadweap [
+    looplist i $playerloadweap [
         if (! (getclientweaphold $ui_hud_focus $i)) [
             _result = 0
         ]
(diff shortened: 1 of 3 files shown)


qreeves:
You don't need to have engine code for this. You can make a persistent string variable with: `defsvarp name value`

nochinator:
should clarify the last commit name, undid changes to client.cpp, nothing else

nochinator:
after further testing it seems like persistent variables made with defsvarp is not avalible in the normal way you get an svar ($varname), doing so throws an error in the console that the variable names alias can't be found. after looking at how other scritps get an svar made in this way it seems like it this is the correct way to get such a variable. it also appears that I am defining it correctly.
defining it before the usage does not change anything.
could someone check my usage?

PR #1568, opened 2024-07-10
Improve crosshair health display

nochinator:
The new implementation has numerous upsides compared to the previous implementation:

- It indicates the player's health more accurately using a color gradient well-known among FPS players. Health above 100 is indicated by tinting the crosshair green.
- It's always visible, even in screenshots. Due to lack of animation, it also avoids hiding the health status on every loop of the animation (since the crosshair became white).

The variable was renamed from `crosshairflash` to `crosshairhealth` to reflect the change.

Fixes #

- [description]


qreeves:
Entire crosshair system is about to be rewritten anyway.

nochinator:
I was trying to merge on my own branch, but accidentaly opened request here, although this is good to know, I won't work on that now

PR #1572, opened 2024-07-14
Update server.cpp to get map from vote starter

### LINKED ISSUE #1012: Problems regarding sendmap/voting a new edit map while online.
While voting custom maps for playing (deathmatch, ctf, etc etc) is fine, it isn't the same for doing that for editing matches.

Have someone vote their map for editing, and the server will only try to grab the map from the player that is there the longest, thus, requiring everyone on the server to just leave the player that has it there, for it to load/vote again, and then the server will have the correct map for editing, which can be quite discouraging.

As for sendmap, it never seems to properly work, as looks like it tries to send the temporary file, mapname_0xblablabla , and the sulfix seems to be causing the issue here, but needs to be tested more, without it, to see if it works or not.

Fixing these would be a godsend, as it would make editing matches way less stressful to play on.

nochinator:
Fixes #1012

- if the server does not have the map of the winning vote then the server now gets the map from the player who started the vote, instead of the player who was on the server the longest. 
- This is implementing by adding an argument to `getmap` for the vote starter's user id (number, not username). it tries to get the map from this player first. it defaults to -1 which is not a player so if it is unspecified then it will use the old functionality.
- The old functionality is used as a fallback in case getting the map from the vote starter fails. If the server already has the map it simply uses that (same as before).

### DIFF
--- a/src/game/server.cpp
+++ b/src/game/server.cpp
@@ -3221,7 +3221,7 @@ namespace server
 
     enum { ALST_TRY = 0, ALST_SPAWN, ALST_SPEC, ALST_EDIT, ALST_WALK, ALST_MAX };
 
-    bool getmap(clientinfo *ci = NULL, bool force = false);
+    bool getmap(clientinfo *ci = NULL, bool force = false, int votingclientnum = -1);
 
     bool crclocked(clientinfo *ci, bool msg = false)
     {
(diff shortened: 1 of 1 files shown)


dc-redeclipse:
Keep in mind that getmap requires the server to be open for no rotational maps voted by the players, means the allowed map list needed to be opened for it.

nochinator:
If I understand what your saying, then this doesn't apply to the changes I've made. It will use normal behavior if the server already has the map (eg. map that is in rotation)
These changes are to fix an issue when a player suggests (on a server that allows out of rotation maps) a map that the server does not have (eg. a edited map or custom map), then the server would get it from the player who was online the longest, rather than the player to start the vote as expected.
If I've misunderstood you're comment then feel free to explain.

PR #1573, opened 2024-07-21
Update servinit.cfg

dc-redeclipse:
add steamserver 0 to reduce problems with not yet active steam ID auth. (should be set 0 in defaults too)


qreeves:
No longer necessary

PR #1575, opened 2024-08-04
Modernize the AppStream metadata

AsciiWolf:
- Update description.
- Add missing name and summary tags.
- Switch to a rDNS app id format that is now required.
- Various fixes to make the file adhere to modern standards.

/cc @qreeves


AsciiWolf:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic

> The ID must follow a reverse-DNS scheme, consisting of {tld}.{vendor}.{product}, for example org.kde.gwenview or com.hugski.colorhug2. Ownership of {vendor}.{tld} in the domain name system guarantees uniqueness of IDs.
> 
> To increase the uniqueness and to distinguish between different pieces of a software suite, it is suggested to append the type name to the component-id in these cases.

(Do not worry about the previous sentence about lowercase letters. Even GNOME uses a mix of lowercase and uppercase for app IDs and there are no problems with it.)

AsciiWolf:
@Calinou Any chance you could take a look at this? Thanks!

AsciiWolf:
@qreeves @Calinou Any chance you could take a look at this? Thanks!

Calinou (approved):

AsciiWolf:
@Calinou Any chance you could also merge this PR? :)

Calinou:
I don't have write access:

image

AsciiWolf:
Oh! I thought that all members of an organization on GitHub (GitLab) should have write access to its repos.

qreeves:
There's only a handful of people with write access to the base repository because it automatically deploys via continuous integration and to prevent people pushing changes without core developer oversight.

I'm currently taking a break from administrative duties and only working on personal tasks unless deemed critical. I'm aware of this ticket, and after taking a look at it I have determined it requires further research and modification by me before being merged. It certainly isn't important enough to be constantly @'ing project moderators.

I am also not happy with the fact that the file contains both static and dynamic information, as the former makes the latter pretty pointless. With the current setup it doesn't even make sense to declare the namespace like that in the Makefile as pretty much nothing else is using that information.

AsciiWolf:
Thanks for the reply!

>I'm aware of this ticket, and after taking a look at it I have determined it requires further research and modification by me before being merged.

Could I ask for some details about this modification that is required? Also, if you need any information about AppStream metadata, I can provide them. I am involved in desktop Linux development and its software distribution (including AppStream and related things) for a quite long time.

>I am also not happy with the fact that the file contains both static and dynamic information, as the former makes the latter pretty pointless. With the current setup it doesn't even make sense to declare the namespace like that in the Makefile as pretty much nothing else is using that information.

I fully agree. These dynamic information are probably not much useful (especially the app id that should be the same as id in the actual metadata file name) and could be all replaced with static ones. If any downstream packaging project wants to modify them, sed or similar tools can be used for that.

AsciiWolf:
@qreeves Any chance you could answer my questions from the previous comment? Thanks!

AsciiWolf:
That's exactly why I asked if they could answer my questions regarding the "required further research and modification". I could provide help.

By the way, I maintain the Xonotic Flatpak on Flathub.

qreeves:
> That's exactly why I asked if they could answer my questions regarding the "required further research and modification". I could provide help.

It means I need to understand the subject better. I can't answer any questions until I do. Please stop @'ing me.

AsciiWolf:
Fair enough. Thanks for the clear answer!

AsciiWolf:
Since this is taking too long, I propose an alternative that is not ideal, but at least will make the AppStream metadata file valid and accepted by modern parsers:

- Do not rename the appdata file or change the app id; do not edit Makefile or system-install.mk.
- Add a `<launchable type="desktop-id">@APPNAME@.desktop</launchable>` tag
- Add the `name`, `developer_name` and `summary` tags from this PR.
- Optionally also add the `provides` tag, new `url` tags and remove empty (none) `content_attribute`s from OARS.

AsciiWolf:
I have made a more simpler, smaller alternative to this PR here: #1599

It does not change the app id or AppStream file name and should be ready to be merged.

AsciiWolf:
Since even the [newer PR](https://github.com/redeclipse/base/pull/1599) that only brings non-invasive changes has not been merged yet, there is no point in keeping this one open.

Closing as replaced by #1599.

PR #1637, opened 2025-04-30
Remove PrefersNonDefaultGPU from desktop file

Jan200101:
PrefersNonDefaultGPU is currently very broken, see https://github.com/ValveSoftware/steam-for-linux/issues/9940

initially added by #1052

### DIFF
--- a/src/install/nix/redeclipse.desktop.am
+++ b/src/install/nix/redeclipse.desktop.am
@@ -19,6 +19,5 @@ Comment[sv]=Förstapersonsskjutare med rörlig spelstil och inbyggd baneditor
 Comment[fr]=Jeu de tir à la première personne avec un éditeur de cartes intégré
 Icon=@APPNAME@
 Exec=@APPNAME@
-PrefersNonDefaultGPU=true
 Categories=Game;ActionGame;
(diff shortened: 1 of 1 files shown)


qreeves:
I'll note that on Windows, this is accomplished with a DLL export that is specifically about NVIDIA Optimus/AMD Express:

```C++
#ifdef WIN32
// Force Optimus setups to use the NVIDIA GPU
extern "C"
{
#ifdef __GNUC__
__attribute__((dllexport))
#else
__declspec(dllexport)
#endif
    DWORD NvOptimusEnablement = 1;

#ifdef __GNUC__
__attribute__((dllexport))
#else
__declspec(dllexport)
#endif
    DWORD AmdPowerXpressRequestHighPerformance = 1;
}
#endif
```

Calinou:
> @Calinou You're probably more aware about the specifications than I, care to weigh in on this?

Yes, I agree it's best to remove it for now.

Linux does not really have a good automatic solution for this just yet. The best you can do is integrate with [GameMode](https://github.com/FeralInteractive/gamemode), but this is not installed by default on most distributions.

PR #1651, opened 2025-07-01
Add .am file for genkey and tessfont binaries

user1-github:
Rationale for this change:

If Linux packages of RE will want to ship the genkey and tessfont binaries, this will make it possible to launch them via a single command, without the need to manually type the full path to the binary in order to launch it.


qreeves:
It's not maintained and we don't officially support installation system-wide. I kinda just take what I get here.

PR #1695, opened 2026-01-28
Steam Input API core support

RobertCochran:
WIP for Steam Input support, as requested on Discord.

There are several things that don't work, aren't implemented the way I'm supposed to, etc.

Perhaps the most important thing to note is that I broke the build for the Red Eclipse server. That needs addressing but it's not short-term important.

### DIFF
--- a/src/game/controller.cpp
+++ b/src/game/controller.cpp
@@ -77,6 +77,9 @@ enum siapi_keycodes {
     SIAPI_MENU_SCROLL_DOWN = -5, // scroll wheel +Y
 };
 
+// For returning SIAPI glyphs as a list
+vector<textkey *> textkeyvec;
+
 // This current controller implementation depends on Steam Input and is not
 // available outside of Steam
 #if defined(USE_STEAM)
@@ -203,9 +206,6 @@ class digital_action_state menu_scroll_up;
 class digital_action_state menu_scroll_down;
 
 DEF_ACTION_SET(EditingControls);
-
-// For returning SIAPI glyphs as a list
-vector<textkey *> textkeyvec;
 
 void init_siapi_handles()
 {
@@ -626,7 +626,7 @@ bool is_siapi_textkey(const char *str)
 
 vector <textkey *> get_siapi_textkeys(const char *str)
 {
-	return NULL;
+	return textkeyvec;
 }
 
 ICOMMAND(0, showsiapibindpanel, "", (), { return; });


src/engine/rendertext.cpp
```diff
@@ -742,22 +767,12 @@ void text_boundsf(const char *str, float &width, float &height, float xpad, floa
...
-    char *name, *file;
-    Texture *tex;
-    textkey() : name(NULL), file(NULL), tex(NULL) {}
-    textkey(char *n, char *f, Texture *t) : name(newstring(n)), file(newstring(f)), tex(t) {}
-    ~textkey()
-    {
-        DELETEA(name);
-        DELETEA(file);
-    }
-};
```
RobertCochran: This moved to `rendertext.h` because I needed the struct to be available elsewhere.

src/game/controller.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+class digital_action_state
+{
+	bool input_last_frame = false;
+	bool input_this_frame = false;
+
+public:
+	InputDigitalActionHandle_t handle = -1;
+	int keymap_id = -1;
+	// Should be more than one origin eventually!
```
RobertCochran: SIAPI actions can have up to 8 origins (which is SIAPIese for 'the button this action is bound to'), but this implementation doesn't yet support actually using them all. It's going to require a non-trivial refactoring of the textkey rendering system to support this.

src/engine/cdpi.cpp
```diff
@@ -40,6 +41,7 @@ namespace cdpi
         ISteamUserStats *stats = NULL;
         ISteamClient *client = NULL, *sclient = NULL;
         ISteamGameServer *serv = NULL;
+	ISteamInput *input = NULL;
```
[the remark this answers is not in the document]
RobertCochran: I fixed everything that wasn't right to the best of my knowledge.

src/game/game.h
```diff
@@ -1246,9 +1246,13 @@ template<class T> inline void flashcolourf(T &r, T &g, T &b, T &f, T br, T bg, T
     f += (bf-f)*amt;
 }
 
+#define mousesens(a,b,c) ((float(a)/float(b))*c)
```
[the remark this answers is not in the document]
RobertCochran: I ended up undoing this at a future point.

bin/amd64/game_actions_967460.vdf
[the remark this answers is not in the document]
RobertCochran: This is now basically complete for this MVP.

src/engine/rendertext.cpp
```diff
@@ -742,37 +767,53 @@ void text_boundsf(const char *str, float &width, float &height, float xpad, floa
...
+    textkey *tk = findtextkey_common(str, textkeys, NULL);
+
+    // Should probably just arrange to have this vector be 1 long at
+    // initialization, but I don't know how to do this...
+    if(!_findtextkeys_container.capacity()) _findtextkeys_container.add(tk);
+    else _findtextkeys_container[0] = tk;
+
+    return _findtextkeys_container;
+
+}
```
RobertCochran: This function was redone to return a vector of textkeys instead of a single textkey. This is necessary because Steam Input allows a single action to have multiple origins, so that codepath could potentially have more than one textkey it needs to return. The pure KB/M path only uses its vector as a wrapper and will only ever return 1 textkey.

src/engine/ui.cpp
```diff
@@ -2057,6 +2057,26 @@ namespace UI
...
 
+        bool menuisgameplay()
+        {
+            if(surfaceinput == type) loopwindows(w,
+            {
+                if(!w->visible || !checkexclusive(w) || w->winstyle >= WINSTYLE_CROSSHAIR) continue;
+                if(!(w->state&STATE_HIDDEN) && (w->winstyle == WINSTYLE_PIE || w->winstyle == WINSTYLE_GAMEPLAY)) return true;
+            });
+            return false;
+        }
```
RobertCochran: This function and its sibling are possibly too dumb, but any more complicated logic and they would just not work correctly. This implementation is good enough for my needs, but making it smarter would be good.

src/game/controller.cpp
```diff
@@ -0,0 +1,634 @@
...
+    InputDigitalActionData_t data = cdpi::steam::input->GetDigitalActionData(controllers[controlleridx], siapi_digital_handle);
+    if(data.bState)
+    {
+        lastinputwassiapi = true;
+        lastusedcontroller = controllers[controlleridx];
+    }
+    return data.bState;
+}
+
+class digital_action_state
```
RobertCochran: Maybe this can be a struct instead?

src/game/controller.cpp
```diff
@@ -0,0 +1,634 @@
...
+	return false;
+}
+
+vector <textkey *> get_siapi_textkeys(const char *str)
+{
+	return textkeyvec;
+}
+
+ICOMMAND(0, showsiapibindpanel, "", (), { return; });
+#endif /* defined(USE_STEAM) */
```
RobertCochran: There are suitable dummy functions exported in non-Steam builds

PR #1708, opened 2026-04-17
add macOS build tools

WhatDidYouExpect:
a persisting bug that goes in and out of fullscreen still exists but if the user manages to tab out at the right time the game is fully playable


q009 (approved):
LGTM

qreeves:
I likely removed the Retina display handling when I removed macOS support. Unfortunately this does not solve the problems that resulted in macOS support being dropped in the first place (code signing & paid developer subscription requirement, OpenGL deprecation). Steam can no longer run unsigned macOS apps as of April 2024.

